PR #2, opened 2019-05-06
Resource API should only support (path, offset) tuple list

aabawer:
Resource API should only support (path, offset)tuple list

As stated in sanlock help, resource API only accepts [(path,offset),...] input form as resource argument.
In this patch a handle in resource parsing is added and a test for invalid input as well.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -33,6 +33,7 @@
 /* Functions prototypes */
 static void __set_exception(int en, char *msg) __sets_exception;
 static int __parse_resource(PyObject *obj, struct sanlk_resource **res_ret) __neg_sets_exception;
+static void util_set_value_error(const char* format, PyObject* obj);
 
 /* Sanlock module */
 PyDoc_STRVAR(pydoc_sanlock, "\
@@ -109,8 +110,10 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
                 __set_exception(EINVAL, "Invalid resource offset");
                 goto exit_fail;
             }
-        } else if (PyString_Check(tuple)) {
-            p = PyString_AsString(tuple);
+        } else {
+            /* handle invalid non-tuple resource input */
+            util_set_value_error("invalid disk value: %s", tuple);
+            goto exit_fail;
         }
 
         if (p == NULL) {
@@ -182,6 +185,17 @@ add_align_flag(long align, uint32_t *flags)
 	return -1;
     }
     return 0;
+}
+
+static void
+util_set_value_error(const char* format, PyObject* obj)
+{
+    const char* str_rep = "";
+    PyObject* rep = PyObject_Repr(obj);
+    if (rep)
+        str_rep = PyString_AsString(rep);
+    PyErr_Format(PyExc_ValueError, format, str_rep);
+    Py_XDECREF(rep);
 }
 
 static PyObject *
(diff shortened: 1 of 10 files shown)


python/sanlock.c
```diff
@@ -33,6 +33,7 @@
 /* Functions prototypes */
 static void __set_exception(int en, char *msg) __sets_exception;
 static int __parse_resource(PyObject *obj, struct sanlk_resource **res_ret) __neg_sets_exception;
+static void __set_value_error(const char* msg, PyObject* obj);
```
nirs: Underscore prefix is reserved for the compiler and standard library, we should not use them.

Instead of adding bad code like the existing one, lets:
- first fix the bad names in the current code, removing the underscore prefix
- add proper name for the new helper

python/sanlock.c
```diff
@@ -109,8 +110,10 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
                 __set_exception(EINVAL, "Invalid resource offset");
                 goto exit_fail;
             }
-        } else if (PyString_Check(tuple)) {
-            p = PyString_AsString(tuple);
+        } else {
+            /* handle invalid non-tuple resource input */
+            __set_value_error("invalid disk value: %s", tuple);
+            goto exit_fail;
```
nirs, line 116: I hoped that we can simplify the if block, but having a minimal change is nice.

python/sanlock.c
```diff
@@ -178,6 +181,17 @@ add_align_flag(long align, uint32_t *flags)
     return 0;
 }
 
+static void
+__set_value_error(const char* msg, PyObject* obj)
```
nirs: "msg" is better named "fmt", since it is a format string, and it should contain a placeholder for obj.

tests/python_test.py
```diff
@@ -284,3 +284,11 @@ def test_write_resource_invalid_align_sector(
     with pytest.raises(ValueError):
         sanlock.write_resource(
             "ls_name", "res_name", disks, align=align, sector=sector)
+
+
+def test_write_resource_invalid_path(tmpdir, sanlock_daemon):
+    path = str(tmpdir.join("resources"))
+    disks = [path]
+    # test parsing resource which is not a list of tuples
```
nirs: The comment should be at the top of the function body.

tests/python_test.py
```diff
@@ -284,3 +284,11 @@ def test_write_resource_invalid_align_sector(
     with pytest.raises(ValueError):
         sanlock.write_resource(
             "ls_name", "res_name", disks, align=align, sector=sector)
+
+
+def test_write_resource_invalid_path(tmpdir, sanlock_daemon):
+    path = str(tmpdir.join("resources"))
+    disks = [path]
```
nirs: We don't need a real path for this test, so we can do:

    disk = ["not a tuple"]

tests/python_test.py
```diff
@@ -284,3 +284,11 @@ def test_write_resource_invalid_align_sector(
...
+    # test parsing resource which is not a list of tuples
+    with pytest.raises(ValueError):
+        sanlock.write_resource("ls_name", "res_name", disks)
```
nirs: We want to check that the error message includes our bad input:

    assert repr("not a tuple") in e.value

Finally, we need to check that this handles correctly non-asci unicode and encoded strings. We can use parameters to run this tests with many bad values.

Also note that a newline is missing at the end of the file.

nirs (changes requested):

python/sanlock.c
aabawer: suggesting "util_" prefix.

aabawer: one trouble at a time :)

aabawer:
will submit new PR for nirs/py3

PR #3, opened 2019-05-06
Module init refactoring

aabawer:
This change refactors sanlock's module init to use inner function. it will serve Python 3 porting in future patches.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -30,6 +30,8 @@
 #define __neg_sets_exception
 #endif
 
+#define MODULE_NAME "sanlock"
+
 /* Functions prototypes */
 static void __set_exception(int en, char *msg) __sets_exception;
 static int __parse_resource(PyObject *obj, struct sanlk_resource **res_ret) __neg_sets_exception;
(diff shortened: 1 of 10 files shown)


python/sanlock.c
```diff
@@ -1674,33 +1676,28 @@ initexception(void)
     return excp;
 }
 
-PyMODINIT_FUNC
-initsanlock(void)
+static int
+__module_init(PyObject* m)
```
nirs: No more underscore prefix please.

python/sanlock.c
```diff
@@ -1693,12 +1691,10 @@ __module_init(PyObject* m)
...
-#define PYSNLK_INIT_ADD_CONSTANT(x, y) \
-    if ((sk_constant = PyInt_FromLong(x)) != NULL) { \
-        if (PyModule_AddObject(m, y, sk_constant)) { \
-            Py_DECREF(sk_constant); \
-            return -1; \
-        } \
+#define PYSNLK_INIT_ADD_CONSTANT(value, label) \
+    if (PyModule_AddIntConstant(m, label, value) != 0) {\
+        Py_DECREF(py_exception); \
+        return -1; \
```
nirs: Please don't add macro with return. This is very confusing to use. Flow should never be hidden in a macro.

python/sanlock.c
```diff
@@ -1540,7 +1540,7 @@ Flags\n\
                             next renewal, ending this event.\n\
   SETEV_REPLACE_EVENT       Replace the existing event/data values of the\n\
                             current event. Without this flag, the operation\n\
-                            will raise SanlockException with -EBUSY error.\n\
+                            will raise sanlockException with -EBUSY error.\n\
```
nirs: If you want to improve the docstrings, you can replace SanlockException with sanlock.SanlockException.

tests/python_test.py
```diff
@@ -198,7 +198,7 @@ def test_acquire_release_resource(tmpdir, sanlock_daemon, size, offset):
     sanlock.add_lockspace("ls_name", 1, ls_path, offset=offset, iotimeout=1)
 
     # Host status is not available until the first renewal.
-    with pytest.raises(sanlock.SanlockException) as e:
+    with pytest.raises(sanlock.sanlockException) as e:
```
nirs: This is wrong.

python/sanlock.c
```diff
@@ -1686,7 +1686,7 @@ __module_init(PyObject* m)
     }
 
     Py_INCREF(py_exception);
-    if (PyModule_AddObject(m, "SanlockException", py_exception)) {
+    if (PyModule_AddObject(m, "sanlockException", py_exception)) {
```
nirs: This is an API change that will break most users of sanlock. We can never change the exception name.

Think about all the code in vdsm doing:

    try:
        ...
    except sanlock.SanlockExcpetion:
        ...

Will fail with NameError when sanlock.SanlockException is raised, hiding the original error from sanlock.

nirs (changes requested):

python/sanlock.c
aabawer: suggesting the "util_" prefix for internal functions.

python/sanlock.c
aabawer: okay, the rest of the function code has the form
AddConstXXX(..)
AddConstXXX(..)
AddConstXXX(..)
etc.
dont we want to check if some const add failed and fail the entire function in this case?
it could save typing multiple ifs.

python/sanlock.c
aabawer: will revert the sanlock exception name changes. wasn't the original intention to do that anyway.

tests/python_test.py
aabawer: will revert

python/sanlock.c
aabawer: will revert

aabawer:
will submit new PR based on nirs/py3

PR #4, opened 2019-05-07
File size constants

barpavel:
1) Adding constants for Kibibyte, Mebibyte, Gibibyte & Tebibyte and using them everywhere in tests.
2) Removing the "import *" style of "constants.py" import in "tests/daemon_test.py".

### DIFF
--- a/tests/constants.py
+++ b/tests/constants.py
@@ -17,3 +17,9 @@ RINDEX_DISK_MAGIC = 0x01042018
 
 RINDEX_ENTRY_SIZE = 64
 RINDEX_ENTRIES_SECTORS = 2000
+
+# A few file/disk size constants.
+KiB = 1024      # 1 Kibibyte = 2^10 bytes = 1024 bytes
+MiB = 1024**2   # 1 Mebibyte = 2^20 bytes = 1,048,576 bytes = 1024 kibibytes
+GiB = 1024**3   # 1 Gibibyte = 2^30 bytes = 1,073,741,824 bytes = 1024 mebibytes
+TiB = 1024**4   # 1 Tebibyte = 2^40 bytes = 1,099,511,627,776 bytes = 1024 gibibytes
--- a/tests/daemon_test.py
+++ b/tests/daemon_test.py
@@ -44,7 +44,7 @@ def test_client_failure():
 
 def test_init_lockspace(tmpdir, sanlock_daemon):
     path = tmpdir.join("lockspace")
-    size = 1024**2
+    size = constants.MiB
     util.create_file(str(path), size)
 
     lockspace = "name:1:%s:0" % path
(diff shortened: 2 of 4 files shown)


tests/python_test.py
```diff
@@ -271,7 +271,7 @@ def test_write_lockspace_invalid_align_sector(
 
 @pytest.mark.parametrize("align, sector", [
     # Invalid alignment
-    (1024, sanlock.SECTOR_SIZE[0]),
+    (constants.KiB, sanlock.SECTOR_SIZE[0]),
     # Invalid sector size
     (sanlock.ALIGN_SIZE[0], 8192),
```
aabawer, line 276: 8*constants.KiB

tests/python_test.py
```diff
@@ -256,7 +256,7 @@ def test_acquire_release_resource(tmpdir, sanlock_daemon, size, offset):
...
```
aabawer, line 261: 8*constants.KiB

tests/daemon_test.py
```diff
@@ -8,7 +8,7 @@
 
 import pytest
 
-from . constants import *
+from . import constants
```
aabawer, line 11: could import as alias:
from . import constants as const

then use the short name:
const.DETLA_DISK_MAGIC
etc.

tests/constants.py
```diff
@@ -17,3 +17,9 @@
 
 RINDEX_ENTRY_SIZE = 64
 RINDEX_ENTRIES_SECTORS = 2000
+
+# A few file/disk size constants.
+KiB = 1024      # 1 Kibibyte = 2^10 bytes = 1024 bytes
+MiB = 1024**2   # 1 Mebibyte = 2^20 bytes = 1,048,576 bytes = 1024 kibibytes
+GiB = 1024**3   # 1 Gibibyte = 2^30 bytes = 1,073,741,824 bytes = 1024 mebibytes
+TiB = 1024**4   # 1 Tebibyte = 2^40 bytes = 1,099,511,627,776 bytes = 1024 gibibytes
```
nirs, line 25: Looks good, but I don't we need the comments about the values.

I think it would be nicer to add the units in a new units.py module - since we really want to write code like:

    from units import *
    ...
    size = 2 * GiB

Interesting that there is a patch in qemu doing the same thing:
http://lists.nongnu.org/archive/html/qemu-block/2019-05/msg00425.html

tests/python_test.py
nirs: 8 * constants.KiB is hard to read, 8192 is better. But if we import the units, 8 * KiB is little nicer.

tests/daemon_test.py
nirs: We have only few constants and their names is pretty clear, so this change does not improve the code.

Lets leave the constants as they are used today.

nirs (changes requested):

barpavel:
Created a clean new PR that incorporates all the above comments:
https://github.com/nirs/sanlock/pull/9

PR #6, opened 2019-05-12
Resource API should only support tuple list for resource argument

aabawer:
Added code to sanlock module to validate and error. Added test.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -33,6 +33,7 @@
 /* Functions prototypes */
 static void __set_exception(int en, char *msg) __sets_exception;
 static int __parse_resource(PyObject *obj, struct sanlk_resource **res_ret) __neg_sets_exception;
+static void set_value_error(const char* format, PyObject* obj);
 
 /* Sanlock module */
 PyDoc_STRVAR(pydoc_sanlock, "\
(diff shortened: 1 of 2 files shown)


python/sanlock.c
```diff
@@ -184,6 +187,17 @@ add_align_flag(long align, uint32_t *flags)
     return 0;
 }
 
+static void
+util_set_value_error(const char* format, PyObject* obj)
```
nirs: Why we need the util_ prefix?

set_value_error() is a static function and It should not conflict with any other api we use.

tests/python_test.py
```diff
@@ -285,3 +285,12 @@ def test_write_resource_invalid_align_sector(
...
         sanlock.write_resource(
             "ls_name", "res_name", disks, align=align, sector=sector)
+
+
+def test_write_resource_invalid_path(tmpdir, sanlock_daemon):
+    # Test parsing a resource which is not a list of tuples
+    disks = ["invalid resource tuple"]
+    with pytest.raises(ValueError) as e:
+        sanlock.write_resource("ls_name", "res_name", disks)
+    assert "invalid resource tuple" in repr(e.value)
```
nirs: We want to test that the result of repr("invalid resource tuple"), which is 'invalid resource tuple' (not the single quote) is inside the error message.

This should be correct:

    invalid_tuple = "invalid"
    disks = [invalid_tuple]
    ...
    assert repr(invalid_tuple) in str(e.value)

Also what about checking non ascii value? This is likely to break with the current code, using:

    str_rep = PyString_AsString(rep);

nirs (changes requested):

tests/python_test.py
aabawer: I think this PR should have two parts:
One with all pushed CR fixes and tested with python 2.
Second part with fixes for python 3:
1. PyString_AsString is replaced with PyUnicode_AsUTF8 in python 3 porting
2. We will be able to test different invalid resource path scenarios over py3 and py2

nirs (approved):

PR #7, opened 2019-05-12
sanlock: Provide internal common module init function

aabawer:
This will serve python 3 porting in future patches to come.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -30,6 +30,8 @@
 #define __neg_sets_exception
 #endif
 
+#define MODULE_NAME "sanlock"
+
 /* Functions prototypes */
 static void __set_exception(int en, char *msg) __sets_exception;
 static int __parse_resource(PyObject *obj, struct sanlk_resource **res_ret) __neg_sets_exception;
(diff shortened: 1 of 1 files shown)


python/sanlock.c
```diff
@@ -1674,75 +1676,87 @@ initexception(void)
...
-    }
+    align = Py_BuildValue("llll",
+                        ALIGNMENT_1M, ALIGNMENT_2M, ALIGNMENT_4M, ALIGNMENT_8M);
+    if (!align || PyModule_AddObject(m, "ALIGN_SIZE", align))
+		goto exit_fail;
+
+	return 0;
+
+exit_fail:
+	Py_XDECREF(py_exception);
```
nirs: This may not be correct.

If we failed to add py_exception to the module we should decref it, but If we succeeded,
we should not since the reference is owned by the module dict. When we decref the module
after error the dict will be deleted with all its contents automatically.

python/sanlock.c
```diff
@@ -1674,75 +1676,87 @@ initexception(void)
...
-    PyObject *sector = Py_BuildValue("ii", SECTOR_SIZE_512, SECTOR_SIZE_4K);
-    if (!sector)
-        return;
-    if (PyModule_AddObject(py_module, "SECTOR_SIZE", sector)) {
-        Py_DECREF(sector);
-        return;
-    }
+    sector = Py_BuildValue("ii", SECTOR_SIZE_512, SECTOR_SIZE_4K);
+    if (!sector || PyModule_AddObject(m, "SECTOR_SIZE", sector))
+		goto exit_fail;
```
nirs: This is more complicated, error prone and harder to follow compared with the previous code.

python/sanlock.c
```diff
@@ -1674,75 +1676,87 @@ initexception(void)
...
-
-    if (py_module == NULL)
-        return;
+	PyObject *align = NULL, *sector = NULL;
 
     if (py_exception == NULL) {
         py_exception = initexception();
         if (py_exception == NULL)
-            return;
+			goto exit_fail;
```
nirs: return -1 is better here - we don't have anything to clean up.

python/sanlock.c
```diff
@@ -1674,75 +1676,87 @@ initexception(void)
     return excp;
 }
 
-PyMODINIT_FUNC
-initsanlock(void)
+static int
+util_module_init(PyObject* m)
```
nirs: I don't think we need the util_ namespace for our static helpers.

python/sanlock.c
```diff
@@ -1674,75 +1676,87 @@ initexception(void)
...
+                        ALIGNMENT_1M, ALIGNMENT_2M, ALIGNMENT_4M, ALIGNMENT_8M);
+    if (!align || PyModule_AddObject(m, "ALIGN_SIZE", align))
+		goto exit_fail;
+
+	return 0;
+
+exit_fail:
+	Py_XDECREF(py_exception);
+	Py_XDECREF(sector);
+	Py_XDECREF(align);
```
nirs: Same issue with sector and align.

I think the best way to handle this is to handle cleanup locally like this:

    Py_INCREF(py_exception);
    if (PyModule_AddObject(py_module, "SanlockException", py_exception)) {
        Py_DECREF(py_exception);
        return -1;
    }

    if (PyModule_AddIntConstant(m, "name", value))
		return -1;
    ...

    PyObject *sector = Py_BuildValue("ii", SECTOR_SIZE_512, SECTOR_SIZE_4K);
    if (!sector)
        return -1;
    if (PyModule_AddObject(m, "SECTOR_SIZE", sector)) {
        Py_DECREF(sector);
        return -1;
    }

No extra cleanup is needed, no hidden control flow, simple.

python/sanlock.c
```diff
@@ -1674,75 +1676,87 @@ initexception(void)
...
+	return -1;
+}
+
+PyMODINIT_FUNC
+initsanlock(void)
+{
+    PyObject *m = NULL;
+
+    m = Py_InitModule4(MODULE_NAME,
+                sanlock_methods, pydoc_sanlock, NULL, PYTHON_API_VERSION);
```
nirs: This is a strange indentation. I think we should use:

    m = Py_InitModule4(MODULE_NAME, sanlock_methods, pydoc_sanlock, NULL, PYTHON_API_VERSION);

Or:

    m = Py_InitModule4(MODULE_NAME, sanlock_methods, pydoc_sanlock, NULL,
                       PYTHON_API_VERSION);

Or:

    m = Py_InitModule4(
        MODULE_NAME, sanlock_methods, pydoc_sanlock, NULL, PYTHON_API_VERSION);

Or:

    m = Py_InitModule4(
        MODULE_NAME,
        sanlock_methods,
        pydoc_sanlock,
        NULL,
        PYTHON_API_VERSION);

I think the last one is best when we have many arguments.

I think the bad indentation is caused by using tabs instead of spaces. The sanlock.c module is using spaces, not tabs like rest of sanlock source.

I think we are missing python/.vimrc or some meta comment to setup proper indent in this file.

python/sanlock.c
```diff
@@ -1674,75 +1676,87 @@ initexception(void)
...
+util_module_init(PyObject* m)
 {
-    PyObject *py_module, *sk_constant;
-
-    py_module = Py_InitModule4("sanlock",
-                sanlock_methods, pydoc_sanlock, NULL, PYTHON_API_VERSION);
-
-    if (py_module == NULL)
-        return;
+	PyObject *align = NULL, *sector = NULL;
```
nirs: Moving cleanup to end of function means we have these NULL objects available to entire function, instead of the relevant parts of the code.

nirs (changes requested):

aabawer:
pushed fixes to PR

python/sanlock.c
aabawer: still TODO

python/sanlock.c
```diff
@@ -1688,75 +1690,96 @@ initexception(void)
...
+    if (PyModule_AddIntConstant(m, "SETEV_CLEAR_EVENT", SANLK_SETEV_CLEAR_EVENT))
+        return -1;
+    if (PyModule_AddIntConstant(m, "SETEV_REPLACE_EVENT", SANLK_SETEV_REPLACE_EVENT))
+        return -1;
+    if (PyModule_AddIntConstant(m, "SETEV_ALL_HOSTS", SANLK_SETEV_ALL_HOSTS))
+        return -1;
 
     /* Tuples with supported sector size and alignment values */
-    PyObject *sector = Py_BuildValue("ii", SECTOR_SIZE_512, SECTOR_SIZE_4K);
+    PyObject* sector = Py_BuildValue("ii", SECTOR_SIZE_512, SECTOR_SIZE_4K);
```
nirs: Should be:

    PyObject *sector ...

python/sanlock.c
```diff
@@ -1688,75 +1690,96 @@ initexception(void)
...
-        return;
+        return -1;
     }
 
-    PyObject *align = Py_BuildValue("llll", ALIGNMENT_1M, ALIGNMENT_2M, ALIGNMENT_4M, ALIGNMENT_8M);
-    if (!align)
-      return;
-    if (PyModule_AddObject(py_module, "ALIGN_SIZE", align)) {
+    PyObject* align = Py_BuildValue("llll",
+                        ALIGNMENT_1M, ALIGNMENT_2M, ALIGNMENT_4M, ALIGNMENT_8M);
```
nirs: Should be:

    PyObject *align = Py_BuildValue(
        "llll", ALIGNMENT_1M, ALIGNMENT_2M, ALIGNMENT_4M, ALIGNMENT_8M);
    
I think I already commented about it previously, did you close and open a new PR, loosing comments?

python/sanlock.c
```diff
@@ -1688,75 +1690,96 @@ initexception(void)
...
+    if (!align) {
+        Py_DECREF(sector);
```
nirs: Here sector is owned by the module - PyMoudule_AddObject in line 1748 stole it - no cleanup is needed.

python/sanlock.c
```diff
@@ -1688,75 +1690,96 @@ initexception(void)
...
+        return -1;
+    }
+    if (PyModule_AddObject(m, "ALIGN_SIZE", align)) {
+        Py_DECREF(sector);
```
nirs: Again, no cleanup is needed for sector.

python/sanlock.c
```diff
@@ -1688,75 +1690,96 @@ initexception(void)
...
+initsanlock(void)
+{
+    PyObject *m = NULL;
+
+    m = Py_InitModule4(
+            MODULE_NAME,
+            sanlock_methods,
+            pydoc_sanlock,
+            NULL,
+            PYTHON_API_VERSION);
```
nirs: The block is indented twice. This is vim default but I like to indent once every level.

python/sanlock.c
```diff
@@ -1688,75 +1690,96 @@ initexception(void)
...
+        return -1;
     }
 
+    return 0;
+}
+
+PyMODINIT_FUNC
+initsanlock(void)
+{
+    PyObject *m = NULL;
```
nirs: Initializing to NULL is not needed.

We can do:

    PyObject *m = Py_InitModule4(...

If you hate the uninitialized pointer.

python/sanlock.c
```diff
@@ -1688,75 +1690,96 @@ initexception(void)
...
+            sanlock_methods,
+            pydoc_sanlock,
+            NULL,
+            PYTHON_API_VERSION);
+
+    if (m == NULL)
+        return;
+
+    if (module_init(m))
+        Py_DECREF(m);
```
nirs: Did you test that this does not segfault on errors?

You can fake a failure in module_init() by setting value error. Importing the module should fail with ImportError or ValueError.

nirs (changes requested):

python/sanlock.c
aabawer: Py_InitModule4 returns Borrowed reference so we shouldnt decref it.
it will segfault when module_init(m) != 0

aabawer:
Pushed fixes after recent CR

python/sanlock.c
```diff
@@ -1688,75 +1690,93 @@ initexception(void)
...
+        NULL,
+        PYTHON_API_VERSION);
+
+    if (m == NULL) {
+        PyErr_Format(PyExc_ImportError, "Failed to allocate module: %s", MODULE_NAME);
+        return;
     }
 
+    if(module_init(m))
+        PyErr_Format(PyExc_ImportError, "Failed to initialize module: %s", MODULE_NAME);
```
nirs: If module_init() failed, it already set the error, or a function it called set the error (e.g. MemoryError). This error will hide the real cause of the failure.

We don't have any way to signal errors in python 2 (except setting an error), so we cannot do anything useful when module_init fail, except returning.

Maybe add a comment about that to save time for the next one looking this code.

python/sanlock.c
```diff
@@ -1688,75 +1690,93 @@ initexception(void)
...
+        return -1;
+    }
+
+    return 0;
+}
+
+PyMODINIT_FUNC
+initsanlock(void)
+{
+    PyObject* m = Py_InitModule4(
```
nirs: Why use Py_InitModule4?

The docs say:
Note Most uses of this function should probably be using the Py_InitModule3() instead; only use this if you are sure you need it.

https://docs.python.org/2/c-api/allocation.html#c.Py_InitModule4

Looking in the source reveals that Py_InitModule3 a macro calling Py_InitModule4:

```
125 #define Py_InitModule3(name, methods, doc) \ 
126     Py_InitModule4(name, methods, doc, (PyObject *)NULL, \
127                PYTHON_API_VERSION)
```

So this is correct, but include details we don't care about.

nirs (changes requested):

aabawer:
Pushed fixes. if we could merge it into nirs/py3 soon it would help for other pending PRs. Thanks

nirs:
Thanks, I push this with minor cleanups as 1e00f0de74b57cbfc0f3cac0afb1b11f59592c9a

PR #9, opened 2019-05-12
Adding constants for Kibibyte, Mebibyte, Gibibyte, etc.

barpavel:
Adding constants for Kibibyte, Mebibyte, Gibibyte, Tebibyte & Pebibyte and using them everywhere in tests.

Signed-off-by: Pavel Bar <pbar@redhat.com>

### DIFF
--- a/tests/daemon_test.py
+++ b/tests/daemon_test.py
@@ -11,6 +11,7 @@ import pytest
 
 from . constants import *
 from . import util
+from . units import *
 
 
 def test_single_instance(sanlock_daemon):
@@ -45,7 +46,7 @@ def test_client_failure():
 
 def test_init_lockspace(tmpdir, sanlock_daemon):
     path = tmpdir.join("lockspace")
-    size = 1024**2
+    size = MiB
     util.create_file(str(path), size)
 
     lockspace = "name:1:%s:0" % path
@@ -62,7 +63,7 @@ def test_init_lockspace(tmpdir, sanlock_daemon):
 
 def test_init_resource(tmpdir, sanlock_daemon):
     path = tmpdir.join("resources")
-    size = 1024**2
+    size = MiB
     util.create_file(str(path), size)
 
     resource = "ls_name:res_name:%s:0" % path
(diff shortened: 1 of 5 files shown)


tests/python_test.py
```diff
@@ -256,9 +258,9 @@ def test_acquire_release_resource(tmpdir, sanlock_daemon, size, offset):
 
 @pytest.mark.parametrize("align, sector", [
     # Invalid alignment
-    (1024, sanlock.SECTOR_SIZE[0]),
+    (KiB, sanlock.SECTOR_SIZE[0]),
     # Invalid sector size
-    (sanlock.ALIGN_SIZE[0], 8192),
+    (sanlock.ALIGN_SIZE[0], 8*KiB),
```
nirs: Missing space:

    8 * KiB

tests/units.py
```diff
@@ -0,0 +1,10 @@
+"""
+Constants for file/disk sizes.
+"""
+
+KiB = 1024
+MiB = 1024**2
+GiB = 1024**3
+TiB = 1024**4
+PiB = 1024**5
+EiB = 1024**6
```
nirs: I don't think we use PiB or EiB, and we will probably never use them.

tests/daemon_test.py
```diff
@@ -78,26 +79,26 @@ def test_init_resource(tmpdir, sanlock_daemon):
 
 def test_format(tmpdir, sanlock_daemon):
     path = tmpdir.join("rindex")
-    size = 1024**2 * 3
+    size = MiB * 3
```
nirs: While we fix this we should also change the order to:

    3 * MiB

Which is more natural.

nirs (changes requested):
Thanks, looks much nicer now!

2 issues:
- always put the count before the size (e.g. 3 * MiB instead of MiB * 3)
- keep space around *

barpavel:
1) Removed "EiB" constant, decided to leave "PiB" constant just in case we'll need it in some foreseeable future.
2) Changed the order when multiplying constant:
`MiB * 3 `--> `3 * MiB`
3) Added space around multiplication sign constant when multiplying constant:
`8*KiB` --> `8 * KiB`
Note: I decided to NOT add spaces in 2 places where this would not feel natural.
`def check_guard(path, size, guard=b"X", guard_size=4*KiB)`

nirs (commented):
Looks like duplicate of #11.

Please take the typos commit out of this PR.

barpavel:
Removed the typos commit.

nirs (approved):
Awesome, thanks!

PR #10, opened 2019-05-13
Python 3 porting

aabawer:
This PR contains the patches for Python 3 porting over the already existing ones in nirs/py3

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -58,6 +58,45 @@ PyUnicode_AsUTF8(PyObject* obj)
     return PyString_AsString(obj);
 }
 
+static int
+PyUnicode_FSConverter(PyObject* arg, void* addr)
+{
+    PyObject *output = NULL;
+    Py_ssize_t size;
+    const char *data;
+
+    if (arg == NULL) {
+        PyErr_Format(PyExc_ValueError, "Input parameter is nil");
+        return 0;
+    }
+    if (PyBytes_Check(arg)) {
+        output = arg;
+        Py_INCREF(arg);
+    }
+    else {
+        output = PyUnicode_AsEncodedString(arg, Py_FileSystemDefaultEncoding, NULL);
+        if (!output) {
+            return 0;
+        }
+        if (!PyBytes_Check(output)) {
+            PyErr_Format(PyExc_ValueError, "Invalid encoded bytes");
+            Py_DECREF(output);
+            return 0;
+        }
+    }
+
+    size = PyBytes_GET_SIZE(output);
+    data = PyBytes_AS_STRING(output);
+    if ((size_t)size != strlen(data)) {
+        PyErr_Format(PyExc_ValueError, "Embedded null byte");
+        Py_XDECREF(output);
+        return 0;
+    }
+
+    *(PyObject**)addr = output;
+    return 1;
+}
+
 #endif
 
 static void
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -30,6 +30,8 @@
 #define __neg_sets_exception
 #endif
 
+#define MODULE_NAME "sanlock"
+
 /* Functions prototypes */
 static void __set_exception(int en, char *msg) __sets_exception;
 static int __parse_resource(PyObject *obj, struct sanlk_resource **res_ret) __neg_sets_exception;
@@ -1688,75 +1690,96 @@ exit_fail:
     return excp;
 }
 
-PyMODINIT_FUNC
-initsanlock(void)
+static int
+util_module_init(PyObject* m)
 {
-    PyObject *py_module, *sk_constant;
-
-    py_module = Py_InitModule4("sanlock",
-                sanlock_methods, pydoc_sanlock, NULL, PYTHON_API_VERSION);
-
-    if (py_module == NULL)
-        return;
-
     if (py_exception == NULL) {
         py_exception = initexception();
         if (py_exception == NULL)
-            return;
+            return -1;
     }
 
     Py_INCREF(py_exception);
-    if (PyModule_AddObject(py_module, "SanlockException", py_exception)) {
+    if (PyModule_AddObject(m, "SanlockException", py_exception)) {
         Py_DECREF(py_exception);
-        return;
-    }
-
-#define PYSNLK_INIT_ADD_CONSTANT(x, y) \
-    if ((sk_constant = PyInt_FromLong(x)) != NULL) { \
-        if (PyModule_AddObject(py_module, y, sk_constant)) { \
-            Py_DECREF(sk_constant); \
-        } \
+        return -1;
     }
 
     /* lockspaces list flags */
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_LSF_ADD, "LSFLAG_ADD");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_LSF_REM, "LSFLAG_REM");
+    if (PyModule_AddIntConstant(m, "LSFLAG_ADD", SANLK_LSF_ADD))
+        return -1;
+    if (PyModule_AddIntConstant(m, "LSFLAG_REM", SANLK_LSF_REM))
+        return -1;
 
     /* resource request flags */
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_REQ_FORCE, "REQ_FORCE");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_REQ_GRACEFUL, "REQ_GRACEFUL");
+    if (PyModule_AddIntConstant(m, "REQ_FORCE", SANLK_REQ_FORCE))
+        return -1;
+    if (PyModule_AddIntConstant(m, "REQ_GRACEFUL", SANLK_REQ_GRACEFUL))
+        return -1;
 
     /* hosts list flags */
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_HOST_FREE, "HOST_FREE");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_HOST_LIVE, "HOST_LIVE");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_HOST_FAIL, "HOST_FAIL");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_HOST_DEAD, "HOST_DEAD");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_HOST_UNKNOWN, "HOST_UNKNOWN");
+    if (PyModule_AddIntConstant(m, "HOST_FREE", SANLK_HOST_FREE))
+        return -1;
+    if (PyModule_AddIntConstant(m, "HOST_LIVE", SANLK_HOST_LIVE))
+        return -1;
+    if (PyModule_AddIntConstant(m, "HOST_FAIL", SANLK_HOST_FAIL))
+        return -1;
+    if (PyModule_AddIntConstant(m, "HOST_DEAD", SANLK_HOST_DEAD))
+        return -1;
+    if (PyModule_AddIntConstant(m, "HOST_UNKNOWN", SANLK_HOST_UNKNOWN))
+        return -1;
 
     /* set event flags */
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_SETEV_CUR_GENERATION, "SETEV_CUR_GENERATION");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_SETEV_CLEAR_HOSTID,   "SETEV_CLEAR_HOSTID");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_SETEV_CLEAR_EVENT,    "SETEV_CLEAR_EVENT");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_SETEV_REPLACE_EVENT,  "SETEV_REPLACE_EVENT");
-    PYSNLK_INIT_ADD_CONSTANT(SANLK_SETEV_ALL_HOSTS,      "SETEV_ALL_HOSTS");
-
-#undef PYSNLK_INIT_ADD_CONSTANT
+    if (PyModule_AddIntConstant(m, "SETEV_CUR_GENERATION", SANLK_SETEV_CUR_GENERATION))
+        return -1;
+    if (PyModule_AddIntConstant(m, "SETEV_CLEAR_HOSTID", SANLK_SETEV_CLEAR_HOSTID))
+        return -1;
+    if (PyModule_AddIntConstant(m, "SETEV_CLEAR_EVENT", SANLK_SETEV_CLEAR_EVENT))
+        return -1;
+    if (PyModule_AddIntConstant(m, "SETEV_REPLACE_EVENT", SANLK_SETEV_REPLACE_EVENT))
+        return -1;
+    if (PyModule_AddIntConstant(m, "SETEV_ALL_HOSTS", SANLK_SETEV_ALL_HOSTS))
+        return -1;
 
     /* Tuples with supported sector size and alignment values */
-    PyObject *sector = Py_BuildValue("ii", SECTOR_SIZE_512, SECTOR_SIZE_4K);
+    PyObject* sector = Py_BuildValue("ii", SECTOR_SIZE_512, SECTOR_SIZE_4K);
     if (!sector)
-        return;
-    if (PyModule_AddObject(py_module, "SECTOR_SIZE", sector)) {
+        return -1;
+    if (PyModule_AddObject(m, "SECTOR_SIZE", sector)) {
         Py_DECREF(sector);
-        return;
+        return -1;
     }
 
-    PyObject *align = Py_BuildValue("llll", ALIGNMENT_1M, ALIGNMENT_2M, ALIGNMENT_4M, ALIGNMENT_8M);
-    if (!align)
-      return;
-    if (PyModule_AddObject(py_module, "ALIGN_SIZE", align)) {
+    PyObject* align = Py_BuildValue("llll",
+                        ALIGNMENT_1M, ALIGNMENT_2M, ALIGNMENT_4M, ALIGNMENT_8M);
+    if (!align) {
+        Py_DECREF(sector);
+        return -1;
+    }
+    if (PyModule_AddObject(m, "ALIGN_SIZE", align)) {
+        Py_DECREF(sector);
         Py_DECREF(align);
-	return;
+        return -1;
     }
 
+    return 0;
+}
+
+PyMODINIT_FUNC
+initsanlock(void)
+{
+    PyObject *m = NULL;
+
+    m = Py_InitModule4(
+            MODULE_NAME,
+            sanlock_methods,
+            pydoc_sanlock,
+            NULL,
+            PYTHON_API_VERSION);
+
+    if (m == NULL)
+        return;
+
+    if (util_module_init(m))
+        Py_DECREF(m);
 }
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -438,15 +438,14 @@ static PyObject *
 py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, max_hosts = 0, num_hosts = 0, use_aio = 1;
-    const char *resource;
     struct sanlk_resource *res;
-    PyObject *disks, *lockspace = NULL;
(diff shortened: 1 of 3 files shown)


tests/python_test.py
```diff
@@ -29,15 +29,18 @@
...
-])
-def test_write_lockspace(tmpdir, sanlock_daemon, size, offset):
-    path = str(tmpdir.join("lockspace"))
+TestParams=[
+        #size,offset,filename,encoding
+        (LOCKSPACE_SIZE, 0, u"ascii", None),
+        (LOCKSPACE_SIZE, 0, u"\u05d0", None),
+        (LARGE_FILE_SIZE, LARGE_FILE_SIZE - LOCKSPACE_SIZE, "ascii", None),
+        (LARGE_FILE_SIZE, LARGE_FILE_SIZE - LOCKSPACE_SIZE, u"\u05d0", "utf-8"),
+]
```
nirs, line 60: This is faster, but the difference is very small, so it does not worth the mess of 
mixing unrelated tests.

To make the tests little cleaner and avoid repeating the parameters, we can have:
```
FILE_NAMES = [
    # name, encoding
    (u"ascii", None),
    (u"\u05d0", None),
    ("ascii", None),
    (u"\u05d0", "utf-8"),
]
```
The sizes should not be refactored like this since we test different values
(LOCKSPACE_SIZE, MIN_RES_SIZE). They happen to be the same for
512 bytes sector size, but this is little ugly. It also not required for adding the
file names and encoding tests.

And TestParams has too generic. We don't have any clue what kind of params are tested.

tests/python_test.py
```diff
@@ -260,14 +258,9 @@ def test_read_resource_owners_invalid_align_size(tmpdir, sanlock_daemon):
...
 
-@pytest.mark.parametrize("size,offset", [
-    # Smallest offset.
-    (MIN_RES_SIZE, 0),
-    # Large offset.
-    (LARGE_FILE_SIZE, LARGE_FILE_SIZE - MIN_RES_SIZE),
-])
-def test_add_rem_lockspace(tmpdir, sanlock_daemon, size, offset):
-    path = str(tmpdir.join("ls_name"))
+@pytest.mark.parametrize("size,offset,filename,encoding", TestParams)
```
nirs: This is slow test, and we don't need to add more permutations here.

If you want to test additional apis with unicode file names add separate fast tests for every api.

It will be nice also to remote the size parameters from this test and test them separately with
other calls, but it is not related to python 3 so lets not spend time on it now.

tests/util.py
```diff
@@ -146,3 +146,10 @@ def check_rindex_entry(entry, name, offset=None, flags=None):
 
     if flags is not None:
         assert e_flags == flags
+
+
+def generate_path(dirname, filename, encoding=None):
+    path = os.path.join(str(dirname), filename)
+    if encoding is not None:
```
nirs, line 153: This is the case where:

    if encoding:
        use it...

Is better, since empty string is not a valid encoding.

tests/util.py
```diff
@@ -146,3 +146,10 @@ def check_rindex_entry(entry, name, offset=None, flags=None):
 
     if flags is not None:
         assert e_flags == flags
+
+
+def generate_path(dirname, filename, encoding=None):
```
nirs, line 151: We need a short docstring explaining what this does.

nirs (changes requested):
Adding failing tests is fine, but you should mark the failing tests with xfail, and include a comment for every failing value, why it fail.

python/sanlock.c
```diff
@@ -1765,6 +1765,36 @@ util_module_init(PyObject* m)
...
 
+#if PY_MAJOR_VERSION >= 3 /* Python 3 module init */
+
+static struct PyModuleDef moduledef = {
+        PyModuleDef_HEAD_INIT,
+        MODULE_NAME,
+        pydoc_sanlock,
+        -1,
+        sanlock_methods,
+};
```
nirs, line 1872: Indented twice instead of once.

nirs (commented):
Looks ready, just minor indentation issue.

python/sanlock.c
```diff
@@ -105,9 +120,15 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
             offset = PyTuple_GetItem(tuple, 1);
 
-            p = PyString_AsString(path);
+            p = PyUnicode_AsUTF8(path);
+
+#if PY_MAJOR_VERSION == 2
+            offset_is_int = PyInt_Check(offset) || PyLong_Check(offset);
+#else
+            offset_is_int = PyLong_Check(offset);
+#endif
```
nirs, line 170: We have only once instance of this, but this is just too ugly. Lets 
have a compatibility wrapper like:

```
   static int
   is_integer(PyObject *obj)
   {
#if PY_MAJOR_VERSION == 2
            return PyInt_Check(obj) || PyLong_Check(obj);
#else
            return PyLong_Check(obj);
#endif
   }
```

Adding this helper and using it is a good patch. We don't want to mix this change with other changes.

python/sanlock.c
```diff
@@ -46,6 +46,20 @@ of the GNU General Public License v2 or (at your option) any later version.");
...
+    Python 2 forward compatibility functions with Python 3
+    TODO: Once Python 2 will reach EOL at 2020 discrad this section
+*/
+#if PY_MAJOR_VERSION == 2
+
+static const char*
+PyUnicode_AsUTF8(PyObject* obj)
+{
+    return PyString_AsString(obj);
+}
```
nirs, line 59: This does not convert to utf8 on python 2.

What is the purpose of this call? Lets find a good function name describing what it does
and add a helper hiding the ifdef - insead of wrapping a helper with the ifdef.

python/sanlock.c
```diff
@@ -91,8 +105,9 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
     res->num_disks = num_disks;
 
     for (i = 0; i < num_disks; i++) {
-        char *p = NULL;
+        const char *p = NULL;
```
nirs: This is not related to the change. If you want improve this do it sedately after we finish with py3 stuff.

python/sanlock.c
```diff
@@ -125,7 +146,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
         if (offset == NULL) {
             res->disks[i].offset = 0;
         } else {
-            res->disks[i].offset = PyInt_AsLong(offset);
+            res->disks[i].offset = PyLong_AsLong(offset);
```
nirs, line 190: Replacing integer with long in places when we can is a great change, please submit it separately.

python/sanlock.c
```diff
@@ -516,15 +537,15 @@ py_read_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
         goto exit_fail;
 
     /* fill the dictionary information: lockspace */
-    if ((ls_entry = PyString_FromString(ls.name)) == NULL)
+    if ((ls_entry = PyBytes_FromString(ls.name)) == NULL)
```
nirs, line 596: Replacing string with bytes is also a great simple change, please submit it separately. I think you can include all instances of lockspace name (read_lockspace) and resource name (read_resource).

python/sanlock.c
```diff
@@ -902,39 +923,39 @@ py_get_lockspaces(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+        if ((ls_value = PyLong_FromLong(lss[i].host_id)) == NULL)
             goto exit_fail;
         rv = PyDict_SetItemString(ls_entry, "host_id", ls_value);
         Py_DECREF(ls_value);
         if (rv != 0)
             goto exit_fail;
 
         /* fill the dictionary information: path */
-        if ((ls_value = PyString_FromString(lss[i].host_id_disk.path)) == NULL)
+        if ((ls_value = PyUnicode_FromString(lss[i].host_id_disk.path)) == NULL)
```
nirs, line 1013: Handling paths should be in a separate patch.

How can we create unicode if we don't know the encoding of the path? We cannot assume that the default encoding will work, since in python 2 it is ascii, and in python 3 utf-8.

We must use here the filesystemencoding - you used it in the python FSConverter.

python/sanlock.c
```diff
@@ -201,39 +222,39 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
             goto exit_fail;
 
         /* fill the dictionary information: host_id */
-        if ((ls_value = PyInt_FromLong(hss[i].host_id)) == NULL)
+        if ((ls_value = PyLong_FromLong(hss[i].host_id)) == NULL)
```
nirs, line 266: From src/sanlock.h:
```
163 struct sanlk_host {                                                                                                                                                      
164         uint64_t host_id;
165         uint64_t generation;
166         uint64_t timestamp;
167         uint32_t io_timeout;
168         uint32_t flags;
169 };
```
host_id should be converted using:

    PyLong_FromUnsignedLongLong

In practice host_id is never more than 2000, but we want to match the sanlock types.
This issue exist also for rest of the fields.

python/sanlock.c
```diff
@@ -902,39 +923,39 @@ py_get_lockspaces(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+        if ((ls_value = PyBytes_FromString(lss[i].name)) == NULL)
             goto exit_fail;
         rv = PyDict_SetItemString(ls_entry, "lockspace", ls_value);
         Py_DECREF(ls_value);
         if (rv != 0)
             goto exit_fail;
 
         /* fill the dictionary information: host_id */
-        if ((ls_value = PyInt_FromLong(lss[i].host_id)) == NULL)
+        if ((ls_value = PyLong_FromLong(lss[i].host_id)) == NULL)
```
nirs, line 1005: From sanlock.h
```
156 struct sanlk_lockspace {                                                                                                                                                 
157         char name[SANLK_NAME_LEN];
158         uint64_t host_id;
159         uint32_t flags; /* SANLK_LSF_ */
160         struct sanlk_disk host_id_disk;
161 };
```

Please match the types.

python/sanlock.c
```diff
@@ -1172,7 +1193,7 @@ py_request(PyObject *self __unused, PyObject *args, PyObject *keywds)
         flags = SANLK_REQUEST_NEXT_LVER;
     } else {
         res->flags |= SANLK_RES_LVER;
-        res->lver = PyInt_AsUnsignedLongMask(version);
+        res->lver = PyLong_AsUnsignedLongMask(version);
```
nirs, line 1263: Please match the types when initializing struct sanlk_resource.

python/sanlock.c
```diff
@@ -1301,7 +1322,7 @@ py_killpath(PyObject *self __unused, PyObject *args, PyObject *keywds)
         size_t arg_len;
 
         item = PyList_GetItem(argslist, i);
-        p = PyString_AsString(item);
+        p = PyUnicode_AsUTF8(item);
```
nirs, line 1395: How can this work in python 2 or 3? What type is item on python 2? python 3?

This change requires a separate patch, with good documentation on the possible input, and a test with some parameters.

nirs (changes requested):
This is a mix of unrelated changes, please separate to simple patches that do one thing:
- Accept both PyInt and PyLong  for disk offset
- Replace PyInt with PyLong (this can be a large patch, since we have lot of integers, but replacing them is one logical change).
- Replace PyString with PyBytes
- Handle arguments list to killpath (maybe use FSConverter?)

python/sanlock.c
```diff
@@ -58,6 +58,45 @@ PyUnicode_AsUTF8(PyObject* obj)
...
+PyUnicode_FSConverter(PyObject* arg, void* addr)
+{
+    PyObject *output = NULL;
+    Py_ssize_t size;
+    const char *data;
+
+    if (arg == NULL) {
+        PyErr_Format(PyExc_ValueError, "Input parameter is nil");
+        return 0;
+    }
```
nirs, line 71: "nil" is strange, this is not objective C. Why not "null"?

python/sanlock.c
```diff
@@ -58,6 +58,45 @@ PyUnicode_AsUTF8(PyObject* obj)
...
+    if (PyBytes_Check(arg)) {
+        output = arg;
+        Py_INCREF(arg);
+    }
+    else {
```
nirs, line 76: We don't check the type, assuming unicode.

The original code does:

    PyOS_FSPath(arg);

Which accept bytes, unicode, or path like object (object implementing __fspath__()).
anything else will raise TypeError.

So the original code is correct checking for bytes and assuming that the type
will be unicode otherwise.

To convert this to python 2 we need to do:

    if (PyUnicode_Check(arg)) {
        ...
    } else if (PyBytes_Check(arg)) {
        ...
    } else {
       PyErr_SetString(PyExc_TypeError, "argument must be a unicode or bytes");
    }

https://docs.python.org/3/c-api/sys.html#c.PyOS_FSPath

python/sanlock.c
```diff
@@ -58,6 +58,45 @@ PyUnicode_AsUTF8(PyObject* obj)
...
+        output = PyUnicode_AsEncodedString(arg, Py_FileSystemDefaultEncoding, NULL);
+        if (!output) {
+            return 0;
+        }
+        if (!PyBytes_Check(output)) {
+            PyErr_Format(PyExc_ValueError, "Invalid encoded bytes");
```
nirs, line 82: Raising value error does not make sense. There is no way output is not bytes object, and if this happens this is internal python error we need to die. ValueError means the user gave your a bad value.

The original code makes sense:

    assert(PyBytes_Check(output));

Lets use it.

python/sanlock.c
```diff
@@ -58,6 +58,45 @@ PyUnicode_AsUTF8(PyObject* obj)
...
+            Py_DECREF(output);
+            return 0;
+        }
+    }
+
+    size = PyBytes_GET_SIZE(output);
+    data = PyBytes_AS_STRING(output);
+    if ((size_t)size != strlen(data)) {
+        PyErr_Format(PyExc_ValueError, "Embedded null byte");
+        Py_XDECREF(output);
```
nirs, line 92: Why XDECREF? output cannot be NULL here.

python/sanlock.c
```diff
@@ -58,6 +58,45 @@ PyUnicode_AsUTF8(PyObject* obj)
     return PyString_AsString(obj);
 }
 
+static int
+PyUnicode_FSConverter(PyObject* arg, void* addr)
```
nirs, line 62: We need good comment for this helper, specifying memory management rules,
and how this version is different from python 3 version, and link to the doc of
the python 3 version.

If you wrote this based on other  code in python, please add link to the original
code.

Thanks for the link in the commit message, but the link should be in the code
so when we have an issue with this, we have an easy way to find it.

nirs (changes requested):

python/sanlock.c
```diff
@@ -144,22 +144,23 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
+        PyObject *tuple, *path = NULL, *offset = NULL, *p = NULL;
         int offset_is_int = 0;
 
         tuple = PyList_GetItem(obj, i);
 
         if (PyTuple_Check(tuple)) {
             if (PyTuple_Size(tuple) != 2) {
-                __set_exception(EINVAL, "Invalid resource tuple");
+                PyErr_Format(PyExc_ValueError, "Invalid resource tuple size: %ld",
+                            PyTuple_Size(tuple));
```
nirs, line 155: This is a good change but it is not related to deploying PyUnicode_FSConverter.

python/sanlock.c
```diff
@@ -144,22 +144,23 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
                 goto exit_fail;
             }
 
             path = PyTuple_GetItem(tuple, 0);
             offset = PyTuple_GetItem(tuple, 1);
 
-            p = PyUnicode_AsUTF8(path);
+            if(!PyUnicode_FSConverter(path, &p))
+                goto exit_fail;
```
nirs, line 163: Lets avoid the temporary PyUnicode_AsUTF8(path) temporary change from the previous patch, since we actually want the path converter.

python/sanlock.c
```diff
@@ -168,19 +169,20 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
 #endif
 
             if (!offset_is_int) {
-                __set_exception(EINVAL, "Invalid resource offset");
+                PyErr_Format(PyExc_ValueError, "Invalid resource offset");
+                Py_XDECREF(p);
```
nirs, line 173: good change, not related to this patch.

python/sanlock.c
```diff
@@ -168,19 +169,20 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
                 goto exit_fail;
             }
-        } else if (PyString_Check(tuple)) {
-            p = PyString_AsString(tuple);
         }
+        //TODO: handle invalid resource tuple (see resource_api PR)
 
         if (p == NULL) {
-            __set_exception(EINVAL, "Invalid resource path");
+            PyErr_Format(PyExc_ValueError, "Invalid resource path");
```
nirs, line 180: Another unrelated change.

python/sanlock.c
```diff
@@ -168,19 +169,20 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
 #endif
 
             if (!offset_is_int) {
-                __set_exception(EINVAL, "Invalid resource offset");
+                PyErr_Format(PyExc_ValueError, "Invalid resource offset");
+                Py_XDECREF(p);
                 goto exit_fail;
             }
-        } else if (PyString_Check(tuple)) {
-            p = PyString_AsString(tuple);
```
nirs, line 113: why did you change it here? this patch should be rebased on the other change, now in my py3 branch.

python/sanlock.c
```diff
@@ -168,19 +169,20 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
 
         if (p == NULL) {
-            __set_exception(EINVAL, "Invalid resource path");
+            PyErr_Format(PyExc_ValueError, "Invalid resource path");
             goto exit_fail;
         }
 
-        strncpy(res->disks[i].path, p, SANLK_PATH_LEN - 1);
+        strncpy(res->disks[i].path, PyBytes_AsString(p), SANLK_PATH_LEN - 1);
+        Py_DECREF(p);
```
nirs, line 185: With so many unrelated changes, I cannot understand where p is defined.

python/sanlock.c
```diff
@@ -348,16 +350,20 @@ static PyObject *
...
     struct sanlk_disk disk;
 
     /* parse python tuple */
-    if (!PyArg_ParseTuple(args, "s", &path)) {
+    if (!PyArg_ParseTuple(args, "O&", PyUnicode_FSConverter, &path)) {
         return NULL;
     }
 
+    if (!path)
+        return NULL;
```
nirs, line 362: How path can be null? If PyUnicode_FSConverter failed, PyArg_ParseTuple will fail.

https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple

python/sanlock.c
```diff
@@ -1395,15 +1438,20 @@ py_killpath(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     /* configure killpath (gil disabled) */
     Py_BEGIN_ALLOW_THREADS
-    rv = sanlock_killpath(sanlockfd, 0, path, kpargs);
+    rv = sanlock_killpath(sanlockfd, 0, PyBytes_AsString(path), kpargs);
     Py_END_ALLOW_THREADS
+    Py_DECREF(path);
```
nirs, line 1431: Decrefing path both on success and error path means we need to switch to single return at the end
and do the cleanup always in the end.

python/sanlock.c
```diff
@@ -1395,15 +1438,20 @@ py_killpath(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
     if (rv != 0) {
         __set_exception(rv, "Killpath script not configured");
         return NULL;
     }
 
     Py_RETURN_NONE;
+
+exit_fail:
+    Py_XDECREF(path);
+    return NULL;
```
nirs, line 1442: We can use rv here to set the exeption and return the correct value:

    out:
        Py_DECREF(path)  // cannot be null
        if (rv != 0) {
            fail...
        }

        Py_RETURN_NONE

nirs (changes requested):
Lets separate the unrelated changes and leave only the deployment as you promised in the commit message.

aabawer:
closing not to mix up  with newer PR.
we keep this as ref for remaining changes to do.

PR #12, opened 2019-05-16
Adding Python 3  functionality

aabawer:
This PR contains the core patches required for sanlock python 3 compatability.
We still need to add the path converter and handle resource and lockspace names a bytes

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -366,15 +366,14 @@ static PyObject *
 py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, max_hosts = 0, num_hosts = 0, use_aio = 1;
-    const char *resource;
     struct sanlk_resource *res;
-    PyObject *disks, *lockspace;
+    PyObject *disks, *lockspace, *resource;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "max_hosts",
                                 "num_hosts", "use_aio", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|iii",
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SSO!|iii",
         kwlist, &lockspace, &resource, &PyList_Type, &disks, &max_hosts,
         &num_hosts, &use_aio)) {
         return NULL;
@@ -387,7 +386,7 @@ py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     /* prepare sanlock names */
     strncpy(res->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
-    strncpy(res->name, resource, SANLK_NAME_LEN);
+    strncpy(res->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* init sanlock resource (gil disabled) */
     Py_BEGIN_ALLOW_THREADS
@@ -623,16 +622,15 @@ py_write_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, max_hosts = 0, num_hosts = 0, clear = 0, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
-    const char *resource;
     struct sanlk_resource *rs;
-    PyObject *disks, *lockspace;
+    PyObject *disks, *lockspace, *resource;
     uint32_t flags = 0;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "max_hosts",
                                 "num_hosts", "clear", "align", "sector", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|iiili",
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SSO!|iiili",
         kwlist, &lockspace, &resource, &PyList_Type, &disks, &max_hosts,
         &num_hosts, &clear, &align, &sector)) {
         return NULL;
@@ -645,7 +643,7 @@ py_write_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     /* prepare sanlock names */
     strncpy(rs->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
-    strncpy(rs->name, resource, SANLK_NAME_LEN);
+    strncpy(rs->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* set alignment/sector flags */
     if (add_align_flag(align, &rs->flags) == -1)
@@ -966,15 +964,14 @@ static PyObject *
 py_acquire(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, sanlockfd = -1, pid = -1, shared = 0;
-    const char *resource;
     struct sanlk_resource *res;
-    PyObject *disks, *version = Py_None, *lockspace;
+    PyObject *disks, *version = Py_None, *lockspace, *resource;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "slkfd",
                                 "pid", "shared", "version", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|iiiO", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SSO!|iiiO", kwlist,
         &lockspace, &resource, &PyList_Type, &disks, &sanlockfd, &pid,
         &shared, &version)) {
         return NULL;
@@ -993,7 +990,7 @@ py_acquire(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     /* prepare sanlock names */
     strncpy(res->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
-    strncpy(res->name, resource, SANLK_NAME_LEN);
+    strncpy(res->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* prepare sanlock flags */
     if (shared) {
@@ -1038,15 +1035,14 @@ static PyObject *
 py_release(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, sanlockfd = -1, pid = -1;
-    const char *resource;
     struct sanlk_resource *res;
-    PyObject *disks, *lockspace;
+    PyObject *disks, *lockspace, *resource;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "slkfd",
                                 "pid", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|ii", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SSO!|ii", kwlist,
         &lockspace, &resource, &PyList_Type, &disks, &sanlockfd, &pid)) {
         return NULL;
     }
@@ -1058,7 +1054,7 @@ py_release(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     /* prepare sanlock names */
     strncpy(res->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
-    strncpy(res->name, resource, SANLK_NAME_LEN);
+    strncpy(res->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* release sanlock resource (gil disabled) */
     Py_BEGIN_ALLOW_THREADS
@@ -1092,15 +1088,14 @@ static PyObject *
 py_request(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, action = SANLK_REQ_GRACEFUL, flags = 0;
-    const char *resource;
     struct sanlk_resource *res;
-    PyObject *disks, *version = Py_None, *lockspace;
+    PyObject *disks, *version = Py_None, *lockspace, *resource;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "action",
                                 "version", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|iO", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SSO!|iO", kwlist,
         &lockspace, &resource, &PyList_Type, &disks, &action, &version)) {
         return NULL;
     }
@@ -1112,7 +1107,7 @@ py_request(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     /* prepare sanlock names */
     strncpy(res->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
-    strncpy(res->name, resource, SANLK_NAME_LEN);
+    strncpy(res->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* prepare the resource version */
     if (version == Py_None) {
@@ -1161,16 +1156,15 @@ py_read_resource_owners(PyObject *self __unused, PyObject *args, PyObject *keywd
     int rv, hss_count = 0;
     int sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
-    const char *resource;
     struct sanlk_resource *res = NULL;
     struct sanlk_host *hss = NULL;
-    PyObject *disks, *ls_list = NULL, *lockspace;
+    PyObject *disks, *ls_list = NULL, *lockspace, *resource;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "align",
                              "sector", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|li", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SSO!|li", kwlist,
         &lockspace, &resource, &PyList_Type, &disks, &align, &sector)) {
         return NULL;
     }
@@ -1182,7 +1176,7 @@ py_read_resource_owners(PyObject *self __unused, PyObject *args, PyObject *keywd
 
     /* prepare sanlock names */
     strncpy(res->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
-    strncpy(res->name, resource, SANLK_NAME_LEN);
+    strncpy(res->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* set resource alignment and sector flags */
 
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -320,8 +320,9 @@ static PyObject *
 py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, max_hosts = 0, num_hosts = 0, use_aio = 1;
-    const char *lockspace, *path;
+    const char *path;
     struct sanlk_lockspace ls;
+    PyObject *lockspace;
 
     static char *kwlist[] = {"lockspace", "path", "offset",
                                 "max_hosts", "num_hosts", "use_aio", NULL};
@@ -330,14 +331,14 @@ py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ss|kiii", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "Ss|kiii", kwlist,
         &lockspace, &path, &ls.host_id_disk.offset, &max_hosts,
         &num_hosts, &use_aio)) {
         return NULL;
     }
 
     /* prepare sanlock names */
-    strncpy(ls.name, lockspace, SANLK_NAME_LEN);
+    strncpy(ls.name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(ls.host_id_disk.path, path, SANLK_PATH_LEN - 1);
 
     /* init sanlock lockspace (gil disabled) */
@@ -365,15 +366,15 @@ static PyObject *
 py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, max_hosts = 0, num_hosts = 0, use_aio = 1;
-    const char *lockspace, *resource;
+    const char *resource;
     struct sanlk_resource *res;
-    PyObject *disks;
+    PyObject *disks, *lockspace;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "max_hosts",
                                 "num_hosts", "use_aio", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ssO!|iii",
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|iii",
         kwlist, &lockspace, &resource, &PyList_Type, &disks, &max_hosts,
         &num_hosts, &use_aio)) {
         return NULL;
@@ -385,7 +386,7 @@ py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
     }
 
     /* prepare sanlock names */
-    strncpy(res->lockspace_name, lockspace, SANLK_NAME_LEN);
+    strncpy(res->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(res->name, resource, SANLK_NAME_LEN);
 
     /* init sanlock resource (gil disabled) */
@@ -420,8 +421,9 @@ py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     int rv, max_hosts = 0, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
     uint32_t io_timeout = 0;
-    const char *lockspace, *path;
+    const char *path;
     struct sanlk_lockspace ls;
+    PyObject *lockspace;
 
     static char *kwlist[] = {"lockspace", "path", "offset", "max_hosts",
                                 "iotimeout", "align", "sector", NULL};
@@ -430,14 +432,14 @@ py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ss|kiIli", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "Ss|kiIli", kwlist,
         &lockspace, &path, &ls.host_id_disk.offset, &max_hosts,
         &io_timeout, &align, &sector)) {
         return NULL;
     }
 
     /* prepare sanlock names */
-    strncpy(ls.name, lockspace, SANLK_NAME_LEN);
+    strncpy(ls.name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(ls.host_id_disk.path, path, SANLK_PATH_LEN - 1);
 
     /* set alignment/sector flags */
@@ -621,16 +623,16 @@ py_write_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, max_hosts = 0, num_hosts = 0, clear = 0, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
-    const char *lockspace, *resource;
+    const char *resource;
     struct sanlk_resource *rs;
-    PyObject *disks;
+    PyObject *disks, *lockspace;
     uint32_t flags = 0;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "max_hosts",
                                 "num_hosts", "clear", "align", "sector", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ssO!|iiili",
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SsO!|iiili",
         kwlist, &lockspace, &resource, &PyList_Type, &disks, &max_hosts,
         &num_hosts, &clear, &align, &sector)) {
         return NULL;
@@ -642,7 +644,7 @@ py_write_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
     }
 
     /* prepare sanlock names */
-    strncpy(rs->lockspace_name, lockspace, SANLK_NAME_LEN);
+    strncpy(rs->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(rs->name, resource, SANLK_NAME_LEN);
 
     /* set alignment/sector flags */
@@ -687,8 +689,9 @@ py_add_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, async = 0, flags = 0;
     uint32_t iotimeout = 0;
-    const char *lockspace, *path;
+    const char *path;
     struct sanlk_lockspace ls;
+    PyObject *lockspace;
 
     static char *kwlist[] = {"lockspace", "host_id", "path", "offset",
                                 "iotimeout", "async", NULL};
@@ -697,7 +700,7 @@ py_add_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "sks|kIi", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "Sks|kIi", kwlist,
         &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset, &iotimeout,
         &async)) {
         return NULL;
@@ -709,7 +712,7 @@ py_add_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     }
 
     /* prepare sanlock names */
-    strncpy(ls.name, lockspace, SANLK_NAME_LEN);
+    strncpy(ls.name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(ls.host_id_disk.path, path, SANLK_PATH_LEN - 1);
 
     /* add sanlock lockspace (gil disabled) */
@@ -738,8 +741,9 @@ static PyObject *
 py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, waitrs = 0, flags = 0;
-    const char *lockspace, *path;
+    const char *path;
     struct sanlk_lockspace ls;
+    PyObject *lockspace;
 
     static char *kwlist[] = {"lockspace", "host_id", "path", "offset",
                                 "wait", NULL};
@@ -748,7 +752,7 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "sks|ki", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "Sks|ki", kwlist,
         &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset,
         &waitrs)) {
         return NULL;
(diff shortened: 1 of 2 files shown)


tests/python_test.py
```diff
@@ -7,6 +7,7 @@
 import io
 import struct
 import time
+import six
```
nirs: six is 3rd party module, so like to import these in the second group like pytest.

tests/python_test.py
```diff
@@ -7,6 +7,7 @@
...
 
 import pytest
```
nirs: Move six here.

tests/python_test.py
```diff
@@ -114,14 +132,15 @@ def test_read_lockspace_4k_invalid_sector_size(sanlock_daemon, user_4k_path):
...
 @pytest.mark.parametrize("size,offset", [
     # Smallest offset.
     (MIN_RES_SIZE, 0),
     # Large offset.
     (LARGE_FILE_SIZE, LARGE_FILE_SIZE - MIN_RES_SIZE),
 ])
-def test_write_resource(tmpdir, sanlock_daemon, size, offset):
-    path = str(tmpdir.join("resources"))
+def test_write_resource(tmpdir, sanlock_daemon, filename, encoding, size, offset):
+    path = util.generate_path(tmpdir, filename, encoding)
```
nirs: Nice easy to review change, thanks!

tests/util.py
```diff
@@ -146,3 +146,15 @@ def check_rindex_entry(entry, name, offset=None, flags=None):
 
     if flags is not None:
         assert e_flags == flags
+
```
nirs: Missing commit message explaining why this utility is needed.

python/sanlock.c
```diff
@@ -1777,6 +1777,34 @@ module_init(PyObject* m)
     return 0;
```
nirs: Missing commit message explaining the current status. I think we cannot compile
the module yet with python 3, since we use symbols that were removed in python 3.

python/sanlock.c
```diff
@@ -128,7 +128,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
         if (offset == NULL) {
```
nirs: The commit message should explain that:

- In python 2 python code does not care about the type of integers returned from sanlock methods.
- In python 3 there is only one type (int), identical to python 2 long.

python/sanlock.c
```diff
@@ -128,7 +128,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
             res->disks[i].offset = 0;
         } else {
-            res->disks[i].offset = PyInt_AsLong(offset);
+            res->disks[i].offset = PyLong_AsLong(offset);
```
nirs: offset is uint64_t, so PyLong_AsUnsignedLongLong()

python/sanlock.c
```diff
@@ -538,7 +538,7 @@ py_read_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
         goto exit_fail;
 
     /* fill the dictionary information: iotimeout */
-    if ((ls_entry = PyInt_FromLong(io_timeout)) == NULL)
+    if ((ls_entry = PyLong_FromLong(io_timeout)) == NULL)
```
nirs: io_timeout is uint32_t

python/sanlock.c
```diff
@@ -1070,7 +1070,7 @@ py_acquire(PyObject *self __unused, PyObject *args, PyObject *keywds)
     /* prepare the resource version */
     if (version != Py_None) {
         res->flags |= SANLK_RES_LVER;
-        res->lver = PyInt_AsUnsignedLongMask(version);
+        res->lver = PyLong_AsUnsignedLongMask(version);
```
nirs: lver is uint64_t, so it cannot overflow. PyLong_AsUnsignedLongLong() should be correct.

python/sanlock.c
```diff
@@ -1186,7 +1186,7 @@ py_request(PyObject *self __unused, PyObject *args, PyObject *keywds)
         flags = SANLK_REQUEST_NEXT_LVER;
     } else {
         res->flags |= SANLK_RES_LVER;
-        res->lver = PyInt_AsUnsignedLongMask(version);
+        res->lver = PyLong_AsUnsignedLongMask(version);
```
nirs: same, uint64_t

Much easier to review and get right now.

python/sanlock.c
```diff
@@ -81,6 +81,17 @@ is_py_integer(PyObject *obj)
 #endif
 }
 
+/* Note that we must keep the input object alive */
```
nirs: This is not detailed enough, and may lead to incorrect usage. How about:

    Return a NUL-terminated representation of the contents of obj.

    obj must be a string object (py2) or Unicode object (py3), otherwise return NULL
    and raise TypeError.

    The returned pointer refers to the internal buffer of string, not a copy. It must not be
    deallocated, and the object must be kept alive while the pointer is used.

We should also make sure that we actually return NULL and raise TypeError in both py2 and py3.

python/sanlock.c
```diff
@@ -116,7 +127,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
             path = PyTuple_GetItem(tuple, 0);
             offset = PyTuple_GetItem(tuple, 1);
 
-            p = PyString_AsString(path);
+            p = pystring_as_cstring(path);
```
nirs: If this return NULL, the issue is a TypeError.

But in line 142 we have:

    if (p == NULL) {
        __set_exception(EINVAL, "Invalid resource path");
        goto exit_fail;
    }

I think this is leftover from code older code. The only case when p == NULL is we could not get the bytes
from the object, which means the error was already set and we can just return NULL.

This error already exist before this change, so we can fix it later.

python/sanlock.c
```diff
@@ -1325,7 +1336,7 @@ py_killpath(PyObject *self __unused, PyObject *args, PyObject *keywds)
         size_t arg_len;
 
         item = PyList_GetItem(argslist, i);
-        p = PyString_AsString(item);
+        p = pystring_as_cstring(item);
 
         if (p == NULL) {
             __set_exception(EINVAL, "Killpath argument not a string");
```
nirs: Same issue - if p is NULL, we already set TypeError, and the SanlockException raised is hides the real error.

We have 2 instances of this issue, lets fix both of them later.

python/sanlock.c
```diff
@@ -953,7 +953,7 @@ py_get_lockspaces(PyObject *self __unused, PyObject *args, PyObject *keywds)
             goto exit_fail;
 
         /* fill the dictionary information: path */
-        if ((ls_value = PyString_FromString(lss[i].host_id_disk.path)) == NULL)
+        if ((ls_value = PyBytes_FromString(lss[i].host_id_disk.path)) == NULL)
```
nirs: This is a path - does not belong to this patch - probably should move to the patch handling paths.

tests/python_test.py
```diff
@@ -281,9 +283,9 @@ def test_read_resource_owners_invalid_align_size(tmpdir, sanlock_daemon):
 
 @pytest.mark.parametrize("size,offset", [
     # Smallest offset.
-    (MIN_RES_SIZE, 0),
+    (MIN_RES_SIZE, long(0)),
     # Large offset.
-    (LARGE_FILE_SIZE, LARGE_FILE_SIZE - MIN_RES_SIZE),
+    (LARGE_FILE_SIZE, long(LARGE_FILE_SIZE - MIN_RES_SIZE)),
```
nirs: Best keep one of the offset as int, since we test different apis in these tests.

nirs (changes requested):
Great work, thanks!

python/sanlock.c
aabawer: > We should also make sure that we actually return NULL and raise TypeError in both py2 and py3.

This is the documented behavior in case of failure:
py2: https://docs.python.org/2/c-api/string.html#c.PyString_AsString
py3: https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize (PyUnicode_AsUTF8 doc refers to this)

nirs: Since you got the links, lets include them in the comment.

python/sanlock.c
aabawer: probably should be PyLong_AsUnsignedLong(offset)  since "unsigned long long" is not strongly typed on gcc systems and depends on actual register size, as can be seen in /usr/include/bits/types.h:
```
#if __WORDSIZE == 64
typedef signed long int __int64_t;
typedef unsigned long int __uint64_t;
#else
__extension__ typedef signed long long int __int64_t;
__extension__ typedef unsigned long long int __uint64_t;
#endif
```
Also discussed in https://stackoverflow.com/questions/32198368/unsigned-long-long-conflict-with-uint64-t

nirs: I'm not sure what do you mean by "not strongly typed".

Python does not give you a way to specify uint64_t, only PY_LONG_LONG. I think this is the correct type if you want to convert from/to [u]int64_t, ensuring that your code will be portable.
See http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf

If you are not sure about that, why not ask on python mailing?

aabawer: havent got to the mailing yet, but is seems that PyLong_AsUnsignedLongLong(offset) fails and returns ((ULL)(-1)) when offset object is not explicitly PyLong in python2. PyLong_AsUnsignedLong(offset) doesn't fail when this happens. 
Seems the py2 impl has fallback in PyLong_AsUnsignedLong(x) when x is not PyLong but a PyInt [1],  which PyLong_AsUnsignedLongLong does not possess[2].
[1] https://github.com/certik/python-2.7/blob/c360290c3c9e55fbd79d6ceacdfc7cd4f393c1eb/Objects/longobject.c#L397
[2] https://github.com/certik/python-2.7/blob/c360290c3c9e55fbd79d6ceacdfc7cd4f393c1eb/Objects/longobject.c#L973

nirs: So we have 2 options:
- add a helpers that accept both PyInt and PyLong, similar to is_py_integer.
- support only PyInt input on python 2, since we never supported PyLong.

Since our goal is to have python 3 version, and making python 2 version better is non-goal, I think the second option is best.

To keep python 2 version consistent, we can return also PyInt from sanlcok on py2, keeping current behavior.

aabawer: me think that we'll need helper in both cases 
as we need to call PyInt_XXXX in py2 and PyLong_XXXX in py3.

python/sanlock.c
aabawer: no Py{Int,Long}_From{UnsignedInt,Int} invented yet.

python/sanlock.c
aabawer: we'd like py3 and py2 to compile and test at this point so we better let it be.
we'll have FSConverter patch later for paths and we'll deal with resource path as well.

tests/python_test.py
aabawer: following discussion about PyInt/PyLong in prev patch we will abandon this patch.
Py3 is long by default, Py2 will stay with Ints as of no-regression action course.

nirs: I have alternative patch for parsing intput:
https://github.com/nirs/sanlock/commit/d3b7f20589dcdfc895ab4d44d404d12fd7e978bf

If we parse input with "K", we have no problem to accept long values, so there is no problem to return long values.

What do you think?

python/sanlock.c
nirs: We have a patch for handling strings and bytes, this patch is about int/long handling.

python/sanlock.c
nirs: @aabawer, I have alternative patch for parsing intput:
https://github.com/nirs/sanlock/commit/d3b7f20589dcdfc895ab4d44d404d12fd7e978bf

If we parse input with "K", we have no problem to accept long values, so there is no problem to return long values.

What do you think?

python/sanlock.c
aabawer: AFAIK its in this commit - creating PyBytes object out of cstring.
The other commit in this PR regarding cstring is for getting the chars representation  of a Pystring/Pyunicode object, the other way around.

nirs: This handles converting sanlock path (cstring) to python type (unicode).

Using bytes is wrong for path. If you don't wan to mix with the change handling path input, split a path handling path output.

nirs: We can use Py_BuildValue("s", path) to convert the paths until we have a better way.

On py2:
s (string) [char *]
Convert a null-terminated C string to a Python object. If the C string pointer is NULL, None is used.

On py3:
s (str or None) [const char *]
Convert a null-terminated C string to a Python str object using 'utf-8' encoding. If the C string pointer is NULL, None is used.

Which is symmetric to what we accept as input for path.

This can also be used for converting numeric values:

   Py_BuildValue("K", offset);

py2:
K (int) [unsigned long long]
Convert a C unsigned long long to a Python integer object.

py3:
K (long) [unsigned PY_LONG_LONG]
Convert a C unsigned long long to a Python long integer object. Only available on platforms that support unsigned long long.

nirs: Note also the magic {items} format:

{items} (dictionary) [matching-items]
Convert a sequence of C values to a Python dictionary. Each pair of consecutive C values adds one item to the dictionary, serving as key and value, respectively.

Can simplify a lot the code creating dicts and list of dicts. Here is an example from _ssl moduel:

```
return Py_BuildValue("{sisisi}", "x509", x509, "crl", crl, "x509_ca", ca);
```

aabawer:
pushed PR updates:
had some re-factorings & rebasings using the Py_BuildValue to get rid from some Py2 native APIs with their Py3 helpers and make life nicer at general.

python/sanlock.c
aabawer: one less, one left.

python/sanlock.c
```diff
@@ -217,55 +217,24 @@ set_value_error(const char* format, PyObject* obj)
...
-        if (rv != 0)
+        /* fill the dictionary information */
+        ls_entry = Py_BuildValue(
+                "{sKsKsKsIsI}",
+                "host_id", hss[i].host_id,
+                "generation", hss[i].generation,
+                "timestamp", hss[i].timestamp,
+                "io_timeout", hss[i].io_timeout,
+                "flags", hss[i].flags);
+        if (ls_entry == NULL)
```
nirs: Awesome.

Does it really fail on python 3 or the commit message needs an update?

python/sanlock.c
```diff
@@ -890,52 +890,17 @@ py_get_lockspaces(PyObject *self __unused, PyObject *args, PyObject *keywds)
         goto exit_fail;
     }
 
-    /* prepare the dictionary holding the information */
-    if ((ls_list = PyList_New(0)) == NULL)
-        goto exit_fail;
```
nirs: How can we add the list if we don't create it first?

python/sanlock.c
```diff
@@ -508,33 +519,16 @@ py_read_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-    rv = PyDict_SetItemString(ls_info, "iotimeout", ls_entry);
-    Py_DECREF(ls_entry);
-    if (rv != 0)
-        goto exit_fail;
+    /* fill the information dictionary */
+    ls_info = Py_BuildValue(
+            read_lockspace_pyvalue_format(),
+            "lockspace", ls.name,
+            "iotimeout", io_timeout);
```
nirs: Moving the ifdef here will make the code much more clear:

```
#if PY_MAJOR_VERSION == 2
    const char *fmt = "{sssI}";
#else
    const char *fmt = "{sysI}";
#endif

ls_info = Py_BuildValue(fmt, "lockspace", ls.name,  "iotimeout", io_timeout);
```

python/sanlock.c
```diff
@@ -589,31 +599,12 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-    Py_DECREF(rs_entry);
-    if (rv != 0)
-        goto exit_fail;
-
-    /* fill the dictionary information: version */
-    if ((rs_entry = PyLong_FromUnsignedLong(rs->lver)) == NULL)
-        goto exit_fail;
-    rv = PyDict_SetItemString(rs_info, "version", rs_entry);
-    Py_DECREF(rs_entry);
-    if (rv != 0)
```
nirs: Having the format here will be much nicer.

python/sanlock.c
```diff
@@ -1792,4 +1820,7 @@ initsanlock(void)
     module_init(m);
 }
 
+#endif
+
 /* vim: set expandtab shiftwidth=4 tabstop=4 : */
+
```
nirs: Why this empty line?

tests/daemon_test.py
```diff
@@ -11,6 +11,7 @@
```
nirs: Seems that you need to rebase on my py3 branch, since Pavel patches are already merged there.

tests/util.py
```diff
@@ -148,3 +148,15 @@ def check_rindex_entry(entry, name, offset=None, flags=None):
...
+def generate_path(dirname, filename, encoding=None):
+    """
+    Return path by joining dirname and filename.
+    The resulting path is either string if encoding is None,
+    or bytes encoded according to given encoding parameter.
+    """
+    path = os.path.join(str(dirname), filename)
+    if encoding:
+        path = path.encode(encoding)
+    return path
```
nirs: Ready, I'll take this to my py3 branch.

tests/python_test.py
```diff
@@ -116,14 +134,15 @@ def test_read_lockspace_4k_invalid_sector_size(sanlock_daemon, user_4k_path):
...
 @pytest.mark.parametrize("size,offset", [
     # Smallest offset.
     (MIN_RES_SIZE, 0),
     # Large offset.
     (LARGE_FILE_SIZE, LARGE_FILE_SIZE - MIN_RES_SIZE),
 ])
-def test_write_resource(tmpdir, sanlock_daemon, size, offset):
-    path = str(tmpdir.join("resources"))
+def test_write_resource(tmpdir, sanlock_daemon, filename, encoding, size, offset):
+    path = util.generate_path(tmpdir, filename, encoding)
```
nirs: Ready, will take to my py3 branch.

src/direct.c
```diff
@@ -339,7 +339,7 @@ static int do_delta_action(int action,
```
nirs: Needs rebase, already in py3 branch.

python/sanlock.c
nirs: We can eliminate this change by squashing my patch rewriting this function. I'm not fan of keeping temporary changes in the tree.

python/sanlock.c
```diff
@@ -71,6 +71,16 @@ __set_exception(int en, char *msg)
     }
 }
 
+static unsigned long long
+pyinteger_as_ullmask(PyObject *obj)
```
nirs: I like the short notation, but it doe snot work for long (pyinteger_as_l?), so better keep the long name:

    pyinteger_as_unsigned_long_long_mask

python/sanlock.c
```diff
@@ -124,44 +124,26 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
     res->num_disks = num_disks;
```
nirs: This patch runs fine with py3 in my tree, the issue is probably another patch.

python/sanlock.c
```diff
@@ -273,7 +273,7 @@ py_register(PyObject *self __unused, PyObject *args)
         return NULL;
```
nirs: Please describe in more details how this behave in py2 and py3.

nirs (changes requested):
Nice to see so much complex code deleted, adding more functionallity :+1: 

Lets try to remove some temporary changes, replaced by Py_BuildValue in the last patches.

python/sanlock.c
nirs: Removing this change in my tree, all the tests pass both on py2 and py3.

Of course we don't have a test for get_lockspaces - we need to add one before we touch this code.

Here is travis build with this fixed, with 4 python versions:
https://travis-ci.org/nirs/sanlock/builds/534274480

nirs:
General note - we don't have tests for sanlock.get_lockspaces() and sanlock.killpath(). We need to add minimal tests before we modify the code.

get_lockspaces() can be tested by adding a call in the test adding lockspace, once before we add (returns []) and once after we add.

killpath can be tested  in a new tests, registering with sanlock (for the fd), and then calling killpath. This should be fast test and we can use FILENAMES for testing how unicode/bytes arguments are handled.

tests/daemon_test.py
nirs: I merged this to my py3 branch, but then I probably removed it by a bad rebase :-). Will push update py3 version soon.

nirs:
@aabawer, I added the missing patches to my py3 branch, and push it. Please rebase.

python/sanlock.c
```diff
@@ -217,55 +217,24 @@ set_value_error(const char* format, PyObject* obj)
...
-
-        /* fill the dictionary information: flags */
-        if ((ls_value = PyInt_FromLong(hss[i].flags)) == NULL)
-            goto exit_fail;
-        rv = PyDict_SetItemString(ls_entry, "flags", ls_value);
-        Py_DECREF(ls_value);
-        if (rv != 0)
+        /* fill the dictionary information */
+        ls_entry = Py_BuildValue(
+                "{sKsKsKsIsI}",
```
nirs: Doc/extending/extending.rst mention a more readable syntax:

    "{s:K,s:K,s:K,s:I,s:I}"

See https://github.com/python/cpython/blob/2.7/Doc/extending/extending.rst#building-arbitrary-values

Turns out that whitespace is ignored, and this works:

    "{s:K, s:K, s:K, s:I, s:I}"

py2:
https://github.com/python/cpython/blob/951af2d7f140be7beb9cda2bcdd54f820c905e45/Python/modsupport.c#L489

py3.6:
https://github.com/python/cpython/blob/2b9d7abdbd4b41e2c624858f5bc80da59d8a681d/Python/modsupport.c#L414

There is a risk in using poorly documented syntax, so I think it will be good idea to ask about this in python mailing list.

python/sanlock.c
aabawer: but ugly. we'll do it when py2 will be officialy dead.

python/sanlock.c
aabawer: ull is commonly used, just like the printf notations. but if you insist.

python/sanlock.c
aabawer: seems legit (see second from last, taken from https://github.com/python/cpython/blob/master/Doc/extending/extending.rst):
   Py_BuildValue("")                        None
   Py_BuildValue("i", 123)                  123
   Py_BuildValue("iii", 123, 456, 789)      (123, 456, 789)
   Py_BuildValue("s", "hello")              'hello'
   Py_BuildValue("y", "hello")              b'hello'
   Py_BuildValue("ss", "hello", "world")    ('hello', 'world')
   Py_BuildValue("s#", "hello", 4)          'hell'
   Py_BuildValue("y#", "hello", 4)          b'hell'
   Py_BuildValue("()")                      ()
   Py_BuildValue("(i)", 123)                (123,)
   Py_BuildValue("(ii)", 123, 456)          (123, 456)
   Py_BuildValue("(i,i)", 123, 456)         (123, 456)
   Py_BuildValue("[i,i]", 123, 456)         [123, 456]
   Py_BuildValue("{s:i,s:i}",
                 "abc", 123, "def", 456)    {'abc': 123, 'def': 456}
   Py_BuildValue("((ii)(ii)) (ii)",
                 1, 2, 3, 4, 5, 6)          (((1, 2), (3, 4)), (5, 6))

python/sanlock.c
aabawer: see prev reply.

python/sanlock.c
aabawer: probably a glitch from all the rebases...
good point!

python/sanlock.c
aabawer: i re-based it to an earlier point in my branch, to get rid of some py3 helpers.

python/sanlock.c
aabawer: i think we can agree on "{s:K,s:K,s:K,s:I,s:I}" as its part of the doc for now.

python/sanlock.c
aabawer: yes. i intentionaly re-based the dict refactoring changes in my branch and removed some py3 helpers patches expected to be replaced. so at this point there is some py3 compatible functionality missing for build.

python/sanlock.c
```diff
@@ -1792,4 +1820,6 @@ initsanlock(void)
     module_init(m);
 }
 
+#endif
+
 /* vim: set expandtab shiftwidth=4 tabstop=4 : */
```
nirs: Looks fine now.

python/sanlock.c
nirs: We can resolve both by squashing later patches, but if you don't find an easy way we can leave this as is.

python/sanlock.c
```diff
@@ -1208,7 +1218,7 @@ py_request(PyObject *self __unused, PyObject *args, PyObject *keywds)
         flags = SANLK_REQUEST_NEXT_LVER;
     } else {
         res->flags |= SANLK_RES_LVER;
-        res->lver = PyInt_AsUnsignedLongMask(version);
+        res->lver = pyinteger_as_unsigned_long_long_mask(version);
         if (res->lver == -1) {
             __set_exception(EINVAL, "Unable to convert the version value");
             goto exit_fail;
```
nirs: Looks fine.

python/sanlock.c
```diff
@@ -217,55 +217,24 @@ set_value_error(const char* format, PyObject* obj)
...
+                "{s:K,s:K,s:K,s:I,s:I}",
+                "host_id", hss[i].host_id,
+                "generation", hss[i].generation,
+                "timestamp", hss[i].timestamp,
+                "io_timeout", hss[i].io_timeout,
+                "flags", hss[i].flags);
+        if (ls_entry == NULL)
             goto exit_fail;
 
         if (PyList_Append(ls_list, ls_entry) != 0)
```
nirs: Looks fine

python/sanlock.c
```diff
@@ -864,6 +864,16 @@ py_rem_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
 }
 
+static const char*
+getlockspaces_pyvalue_format()
+{
+#if PY_MAJOR_VERSION == 2
+    return "{s:s,s:K,s:s,s:K,s:I}";
+#else
+    return "{s:y,s:K,s:s,s:K,s:I}";
+#endif
```
nirs: Looks fine, but

python/sanlock.c
```diff
@@ -895,47 +905,16 @@ py_get_lockspaces(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-
-        /* fill the dictionary information: flags */
-        if ((ls_value = PyInt_FromLong(lss[i].flags)) == NULL)
-            goto exit_fail;
-        rv = PyDict_SetItemString(ls_entry, "flags", ls_value);
-        Py_DECREF(ls_value);
-        if (rv != 0)
+        /* fill the dictionary information */
+        ls_entry = Py_BuildValue(
+                getlockspaces_pyvalue_format(),
```
nirs: The format should be here. It does not make sense to:
- Move the format string far from the code using it
- Add a function to return the format string, you could use a constant

python/sanlock.c
```diff
@@ -460,6 +460,17 @@ py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
 }
 
+const char*
+read_lockspace_pyvalue_format()
+{
+#if PY_MAJOR_VERSION == 2
+    return "{s:s,s:I}";
+#else
+    return "{s:y,s:I}";
+#endif
```
nirs: Same as the other format "helper"

python/sanlock.c
```diff
@@ -508,33 +519,16 @@ py_read_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-    /* fill the dictionary information: iotimeout */
-    if ((ls_entry = PyInt_FromLong(io_timeout)) == NULL)
-        goto exit_fail;
-    rv = PyDict_SetItemString(ls_info, "iotimeout", ls_entry);
-    Py_DECREF(ls_entry);
-    if (rv != 0)
-        goto exit_fail;
+    /* fill the information dictionary */
+    ls_info = Py_BuildValue(
+            read_lockspace_pyvalue_format(),
```
nirs: The format should be here.

python/sanlock.c
```diff
@@ -306,7 +306,7 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
         return NULL;
     }
 
-    return PyInt_FromLong(rv);
+    return Py_BuildValue("i", rv);
 }
 
 /* init_lockspace */
```
nirs: Looks fine

tests/python_test.py
```diff
@@ -305,13 +305,25 @@ def test_add_rem_lockspace(tmpdir, sanlock_daemon, size, offset):
         "ls_name", 1, path, offset=offset, wait=False)
     assert acquired is True
 
+    lls = sanlock.get_lockspaces()
```
nirs: lls is a bit too cryptic for me. Can you think about a name that will be more clear.

python/sanlock.c
```diff
@@ -1546,22 +1551,22 @@ with -EBUSY error in this case.\n\
...
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "SKKK|KI", kwlist,
         &lockspace, &he.host_id, &he.generation, &he.event, &he.data,
         &flags)) {
         return NULL;
     }
 
     Py_BEGIN_ALLOW_THREADS
-    rv = sanlock_set_event(lockspace, &he, flags);
+    rv = sanlock_set_event(PyBytes_AsString(lockspace), &he, flags);
     Py_END_ALLOW_THREADS
```
nirs, line 1545: We don't have tests for this code, need to add minimal tests.

This patch is rather big, lets defer this work for now. I want to merge the first path, where all the code and tests pass on python 3, and work on better types in python 3 and smarter path handling in the next phase.

nirs (changes requested):
I reviewed up to:
python: Treat lockspace name argument as PyBytes

Lets finish all the patches up to this patch, before we continue with making python 3 API better.

python/sanlock.c
aabawer: later patches only cover the use-case  in __parse_resource out of the three occurrences shown here. so we can leave it as is at this point.

python/sanlock.c
aabawer: i think we came to realize at that point that #ifdefs in middle of existing code are much uglier.
if you find that having the format string outside the Py_BuildValue call is hard to follow, i think we can settle the case by using Py_BuildValue helper per problematic Py3-Py2 format use-case. for example here we could use helper like:
`static PyObject*
getlockspaces_pyvalue(const sanlk_lockspace* lss)
{
/* call the Py_BuildValue per Py Version and return the PyObject */
....
}`

python/sanlock.c
aabawer: see prev reply.

python/sanlock.c
aabawer: replied.

python/sanlock.c
aabawer: replied.

tests/python_test.py
aabawer: ls_list?

python/sanlock.c
aabawer: Okay.

python/sanlock.c
nirs: We discussed avoiding repeated usage of same ifdefs, by creating a wrapper hiding the difference between python 2 and 3.

There is nothing wrong in having one ifdef in the function calling Py_BuildValue. The important thing is to have the format string near the arguments.

If we have several places converting sanlock lockspace to dict, we can add a helper like lockspace_to_dict using Py_BuildValue internally, but if we have only once place, I don't see a need for such helper.

tests/python_test.py
nirs: ls is a shortcut for lockspace, and used in "ls_name". ls_list is not consistent with this usage. How about:

    lockspaces = anlock.get_lockspaces()

This works nicely with existing code, like:

    owners = sanlock.read_resource_owners("ls_name", "res_name", disks)

python/sanlock.c
nirs: Merged into my py3 branch.

tests/util.py
nirs: Merged into my py3 branch.

tests/python_test.py
```diff
@@ -116,14 +134,15 @@ def test_read_lockspace_4k_invalid_sector_size(sanlock_daemon, user_4k_path):
...
     (MIN_RES_SIZE, 0),
     # Large offset.
     (LARGE_FILE_SIZE, LARGE_FILE_SIZE - MIN_RES_SIZE),
 ])
-def test_write_resource(tmpdir, sanlock_daemon, size, offset):
-    path = str(tmpdir.join("resources"))
+def test_write_resource(tmpdir, sanlock_daemon, filename, encoding, size, offset):
+    path = util.generate_path(tmpdir, filename, encoding)
     util.create_file(path, size)
     disks = [(path, offset)]
```
nirs, line 148: Merged into my py3 branch.

python/sanlock.c
nirs: Merged into my py3 branch.

python/sanlock.c
```diff
@@ -71,6 +71,16 @@ __set_exception(int en, char *msg)
     }
 }
 
+static unsigned long long
```
nirs: Fixed to uint64_t, matching sanlock type.

python/sanlock.c
nirs: Merged into my py3 branch.

python/sanlock.c
nirs: Merged into my py3 branch.

python/sanlock.c
```diff
@@ -895,47 +895,20 @@ py_get_lockspaces(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+                "{s:s,s:K,s:s,s:K,s:I}",
+#else
+                "{s:y,s:K,s:s,s:K,s:I}",
+#endif
+                "lockspace", lss[i].name,
+                "host_id", lss[i].host_id,
+                "path", lss[i].host_id_disk.path,
+                "offset", lss[i].host_id_disk.offset,
+                "flags", lss[i].flags);
+        if (ls_entry == NULL)
```
nirs: Much nicer like this, thanks!

python/sanlock.c
```diff
@@ -582,31 +582,16 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+#else
+        "{s:y,s:y,s:K}",
+#endif
+        "lockspace", rs->lockspace_name,
+        "resource", rs->name,
+        "version", rs->lver);
+    if (rs_info  == NULL)
         goto exit_fail;
 
     /* success */
```
nirs: Merged into my py3 branch.

python/sanlock.c
```diff
@@ -508,33 +508,20 @@ py_read_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
     return ls_info;
-
-    /* failure */
-exit_fail:
-    Py_XDECREF(ls_info);
-    return NULL;
 }
 
 /* read_resource */
```
nirs: Merged into my py3 branch.

python/sanlock.c
nirs: Merged into my py3 branch.

python/sanlock.c
nirs: Merged into my py3 branch.

nirs (commented):
I merged everything up to:
https://github.com/nirs/sanlock/pull/12/commits/b15170d81cf1ad660fd2c7d89decdfd87ba81159

Please rebase on my py3 branch.

nirs:
@aabawer please rebase on py3 branch and remove the patches up to 
python: Use Py_BuildValue("i", ...) to create py integers
which were merged into my branch.

Lets continue this work in a new more focused PR for handling lockspace and resource names on python 3.

aabawer:
Rebased.

tests/python_test.py
```diff
@@ -305,13 +305,25 @@ def test_add_rem_lockspace(tmpdir, sanlock_daemon, size, offset):
...
     acquired = sanlock.inq_lockspace(
         "ls_name", 1, path, offset=offset, wait=False)
     assert acquired is False
 
+    lockspaces = sanlock.get_lockspaces()
+    assert lockspaces == []
+
 
 def test_add_rem_lockspace_async(tmpdir, sanlock_daemon):
     path = str(tmpdir.join("ls_name"))
```
aabawer, line 329: not merged into nirs/py3 for a reason?

nirs: I stopped when all the tests passed. This one looks ready now. I think you posted this version while I was merging, and I missed this change.

nirs (commented):
Looks good in quick look. I need more time to review.

I would be nice if you make the changes in this order:
- add missing test for all calls using lockspace/resource names
- change the test to use always bytes for py3
- keep both b"name" and "name" in py2, ensuring that they are really the same.
  current vdsm code uses "name", and sanlock must be compatible with it when running on python 2.
  vdsm for python 3 will have to change all calls to use b"name".
- If some parameters fail with py3, mark the failing parameter as xfail
- make the changes the module to require bytes in py3, removing possible xfails.

Now that the all the tests are passing, posting a patch breaking the tests is not an option.

nirs:
@aabawer can you close this PR and start to a new PR focused on the bytes changes?

aabawer:
Ack.

aabawer:
we'll have another PR for lockspace & resource name bytes

PR #14, opened 2019-05-20
Parse lockspace and resource names as PyBytes

aabawer:
WIP:
-    add missing test for all calls using lockspace/resource names
 -   change the test to use always bytes for py3
-    keep both b"name" and "name" in py2, ensuring that they are really the same.
    current vdsm code uses "name", and sanlock must be compatible with it when running on python 2.
    vdsm for python 3 will have to change all calls to use b"name".
-    If some parameters fail with py3, mark the failing parameter as xfail
 -   make the changes the module to require bytes in py3, removing possible xfails.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -132,14 +132,14 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
 
         if (!PyTuple_Check(disk)) {
             set_error(PyExc_ValueError, "Invalid disk %s", disk);
-            goto exit_fail;
+            goto finally;
 
         }
 
         if (!PyArg_ParseTuple(disk, "sK", &path, &offset)) {
             /* Override the error since it confusing in this context. */
             set_error(PyExc_ValueError, "Invalid disk %s", disk);
-            goto exit_fail;
+            goto finally;
         }
 
         strncpy(res->disks[i].path, path, SANLK_PATH_LEN - 1);
@@ -149,7 +149,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
     *res_ret = res;
     return 0;
 
-exit_fail:
+finally:
     free(res);
     return -1;
 }
@@ -222,7 +222,7 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
 
     /* prepare the dictionary holding the information */
     if ((ls_list = PyList_New(0)) == NULL)
-        goto exit_fail;
+        goto finally;
 
     for (i = 0; i < hss_count; i++) {
 
@@ -235,10 +235,10 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
                 "io_timeout", hss[i].io_timeout,
                 "flags", hss[i].flags);
         if (ls_entry == NULL)
-            goto exit_fail;
+            goto finally;
 
         if (PyList_Append(ls_list, ls_entry) != 0)
-            goto exit_fail;
+            goto finally;
 
         Py_DECREF(ls_entry);
     }
@@ -246,7 +246,7 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
     return ls_list;
 
     /* failure */
-exit_fail:
+finally:
     Py_XDECREF(ls_entry);
     Py_XDECREF(ls_list);
     return NULL;
@@ -309,6 +309,43 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
     return Py_BuildValue("i", rv);
 }
 
+/*
+ * Convert parsed arg into PyBytes object.
+ * For Python 2:
+ * If arg is unicode onject, ascii encode it to new PyBytes object passed by addr.
+ * If arg is a bytes object, inc its refcount and pass it in addr.
+ * Set TypeError and return 0 if arg doens not comply to any of the above.
+ * Return 1 on a successful conversion.
+ * For Python 3:
+ * If arg is a bytes object, inc its refcount and pass it in addr.
+ * Set TypeError and return 0 otherwise.
+ * Return 1 on a successful conversion.
+*/
+static int
+convert_to_pybytes(PyObject* arg, void *addr)
+{
+    assert(arg && "convert_to_pybytes called with NULL arg");
+
+#if PY_MAJOR_VERSION == 2
+    if (PyUnicode_Check(arg)) {
+        PyObject *bytes = PyUnicode_AsASCIIString(arg);
+        if (bytes == NULL)
+            return 0;
+        *(PyObject **)addr = bytes;
+        return 1;
+    }
+#endif
+
+    if (PyBytes_Check(arg)) {
+        Py_INCREF(arg);
+        *(PyObject **)addr = arg;
+        return 1;
+    }
+
+    set_error(PyExc_TypeError, "Argument type is not bytes: %s", arg);
+    return 0;
+}
+
 /* init_lockspace */
 PyDoc_STRVAR(pydoc_init_lockspace, "\
 init_lockspace(lockspace, path, offset=0, max_hosts=0, num_hosts=0, \
@@ -319,8 +356,9 @@ Initialize a device to be used as sanlock lockspace.");
 static PyObject *
 py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, max_hosts = 0, num_hosts = 0, use_aio = 1;
-    const char *lockspace, *path;
+    int rv = -1, max_hosts = 0, num_hosts = 0, use_aio = 1;
+    PyObject *lockspace = NULL;
+    const char *path;
     struct sanlk_lockspace ls;
 
     static char *kwlist[] = {"lockspace", "path", "offset",
@@ -330,14 +368,14 @@ py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ss|kiii", kwlist,
-        &lockspace, &path, &ls.host_id_disk.offset, &max_hosts,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&s|kiii", kwlist,
+        convert_to_pybytes, &lockspace, &path, &ls.host_id_disk.offset, &max_hosts,
         &num_hosts, &use_aio)) {
-        return NULL;
+        goto finally;
     }
 
     /* prepare sanlock names */
-    strncpy(ls.name, lockspace, SANLK_NAME_LEN);
+    strncpy(ls.name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(ls.host_id_disk.path, path, SANLK_PATH_LEN - 1);
 
     /* init sanlock lockspace (gil disabled) */
@@ -347,9 +385,13 @@ py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock lockspace init failure");
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
 }
 
@@ -364,29 +406,28 @@ The disks must be in the format: [(path, offset), ... ]");
 static PyObject *
 py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, max_hosts = 0, num_hosts = 0, use_aio = 1;
-    const char *lockspace, *resource;
-    struct sanlk_resource *res;
-    PyObject *disks;
+    int rv = -1, max_hosts = 0, num_hosts = 0, use_aio = 1;
+    PyObject *disks, *lockspace = NULL, *resource= NULL;
+    struct sanlk_resource *res = NULL;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "max_hosts",
                                 "num_hosts", "use_aio", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ssO!|iii",
-        kwlist, &lockspace, &resource, &PyList_Type, &disks, &max_hosts,
-        &num_hosts, &use_aio)) {
-        return NULL;
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&O&O!|iii",
+        kwlist, convert_to_pybytes, &lockspace, convert_to_pybytes, &resource, 
+        &PyList_Type, &disks, &max_hosts, &num_hosts, &use_aio)) {
+        goto finally;
     }
 
     /* parse and check sanlock resource */
     if (__parse_resource(disks, &res) < 0) {
-        return NULL;
+        goto finally;
     }
 
     /* prepare sanlock names */
-    strncpy(res->lockspace_name, lockspace, SANLK_NAME_LEN);
-    strncpy(res->name, resource, SANLK_NAME_LEN);
+    strncpy(res->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
+    strncpy(res->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* init sanlock resource (gil disabled) */
     Py_BEGIN_ALLOW_THREADS
@@ -395,15 +436,16 @@ py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock resource init failure");
-        goto exit_fail;
+        goto finally;
     }
 
-    free(res);
+finally:
+    Py_XDECREF(lockspace);
+    Py_XDECREF(resource);
+    if (res) free(res);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
-
-exit_fail:
-    free(res);
-    return NULL;
 }
 
 /* write_lockspace */
@@ -417,10 +459,11 @@ Sector can be one of (512, 4096).");
 static PyObject *
 py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, max_hosts = 0, sector = SECTOR_SIZE_512;
+    int rv = -1, max_hosts = 0, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
     uint32_t io_timeout = 0;
-    const char *lockspace, *path;
+    PyObject *lockspace = NULL;
+    const char *path;
     struct sanlk_lockspace ls;
 
     static char *kwlist[] = {"lockspace", "path", "offset", "max_hosts",
@@ -430,22 +473,22 @@ py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ss|kiIli", kwlist,
-        &lockspace, &path, &ls.host_id_disk.offset, &max_hosts,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&s|kiIli", kwlist,
+        convert_to_pybytes, &lockspace, &path, &ls.host_id_disk.offset, &max_hosts,
         &io_timeout, &align, &sector)) {
-        return NULL;
+        goto finally;
     }
 
     /* prepare sanlock names */
-    strncpy(ls.name, lockspace, SANLK_NAME_LEN);
+    strncpy(ls.name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(ls.host_id_disk.path, path, SANLK_PATH_LEN - 1);
 
     /* set alignment/sector flags */
     if (add_align_flag(align, &ls.flags) == -1)
-        return NULL;
+        goto finally;
 
     if (add_sector_flag(sector, &ls.flags) == -1)
-        return NULL;
+        goto finally;
 
     /* write sanlock lockspace (gil disabled) */
     Py_BEGIN_ALLOW_THREADS
@@ -454,9 +497,13 @@ py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock lockspace write failure");
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
 }
 
@@ -558,7 +605,7 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
     /* parse python tuple */
     if (!PyArg_ParseTupleAndKeywords(args, keywds, "s|kli", kwlist,
         &path, &(rs->disks[0].offset), &align, &sector)) {
-        goto exit_fail;
+        goto finally;
     }
 
     /* prepare the resource disk path */
@@ -566,10 +613,10 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     /* set alignment/sector flags */
     if (add_align_flag(align, &rs->flags) == -1)
-        goto exit_fail;
+        goto finally;
 
     if (add_sector_flag(sector, &rs->flags) == -1)
-        goto exit_fail;
+        goto finally;
 
     /* read sanlock resource (gil disabled) */
     Py_BEGIN_ALLOW_THREADS
@@ -578,7 +625,7 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock resource read failure");
-        goto exit_fail;
+        goto finally;
     }
 
     /* prepare the dictionary holding the information */
@@ -592,17 +639,12 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
         "resource", rs->name,
         "version", rs->lver);
     if (rs_info  == NULL)
-        goto exit_fail;
+        goto finally;
 
-    /* success */
-    free(rs);
+finally:
+    if(rs) free(rs);
+    /* rs_info is either NULL or returned - no decref here */
     return rs_info;
-
-    /* failure */
-exit_fail:
-    free(rs);
-    Py_XDECREF(rs_info);
-    return NULL;
 }
 
 /* write_resource */
@@ -619,38 +661,37 @@ Sector can be one of (512, 4096).");
 static PyObject *
 py_write_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, max_hosts = 0, num_hosts = 0, clear = 0, sector = SECTOR_SIZE_512;
+    int rv = -1, max_hosts = 0, num_hosts = 0, clear = 0, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
-    const char *lockspace, *resource;
-    struct sanlk_resource *rs;
-    PyObject *disks;
+    PyObject *disks, *lockspace = NULL, *resource = NULL;
     uint32_t flags = 0;
+    struct sanlk_resource *rs = NULL;
 
     static char *kwlist[] = {"lockspace", "resource", "disks", "max_hosts",
                                 "num_hosts", "clear", "align", "sector", NULL};
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "ssO!|iiili",
-        kwlist, &lockspace, &resource, &PyList_Type, &disks, &max_hosts,
-        &num_hosts, &clear, &align, &sector)) {
-        return NULL;
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&O&O!|iiili",
+        kwlist, convert_to_pybytes, &lockspace, convert_to_pybytes, &resource,
+        &PyList_Type, &disks, &max_hosts, &num_hosts, &clear, &align, &sector)) {
+        goto finally;
     }
 
     /* parse and check sanlock resource */
     if (__parse_resource(disks, &rs) < 0) {
-        return NULL;
+        goto finally;
     }
 
     /* prepare sanlock names */
-    strncpy(rs->lockspace_name, lockspace, SANLK_NAME_LEN);
-    strncpy(rs->name, resource, SANLK_NAME_LEN);
+    strncpy(rs->lockspace_name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
+    strncpy(rs->name, PyBytes_AsString(resource), SANLK_NAME_LEN);
 
     /* set alignment/sector flags */
     if (add_align_flag(align, &rs->flags) == -1)
-        goto exit_fail;
+        goto finally;
 
     if (add_sector_flag(sector, &rs->flags) == -1)
-        goto exit_fail;
+        goto finally;
 
     if (clear) {
         flags |= SANLK_WRITE_CLEAR;
@@ -663,15 +704,17 @@ py_write_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock resource write failure");
-        goto exit_fail;
+        goto finally;
     }
 
-    free(rs);
-    Py_RETURN_NONE;
 
-exit_fail:
-    free(rs);
-    return NULL;
+finally:
+    Py_XDECREF(lockspace);
+    Py_XDECREF(resource);
+    if(rs) free(rs);
+    if (rv != 0)
+        return NULL;
+    Py_RETURN_NONE;
 }
 
 /* add_lockspace */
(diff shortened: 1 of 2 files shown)


tests/python_test.py
```diff
@@ -61,22 +61,22 @@ def test_write_lockspace(tmpdir, sanlock_daemon, filename, encoding, size, offse
     util.create_file(path, size)
 
     # Test read and write with default alignment and sector size values.
-    sanlock.write_lockspace("name", path, offset=offset, iotimeout=1)
+    sanlock.write_lockspace(b"name", path, offset=offset, iotimeout=1)
```
nirs, line 82: lockspace and resource name should be parametric:
- "\xd7\x90" (py2, py3)
- u"ascii" (py2 compatibility)

We can do this for the same tests we use parametric paths.

tests/python_test.py
```diff
@@ -147,7 +147,7 @@ def test_write_resource(tmpdir, sanlock_daemon, filename, encoding, size, offset
     disks = [(path, offset)]
 
     # Test read and write with default alignment and sector size values.
-    sanlock.write_resource("ls_name", "res_name", disks)
+    sanlock.write_resource(b"ls_name", "res_name", disks)
```
nirs: Same parametric approach for resource names.

python/sanlock.c
```diff
@@ -320,8 +320,9 @@ static PyObject *
 py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv, max_hosts = 0, num_hosts = 0, use_aio = 1;
-    const char *lockspace, *path;
+    const char *path;
     struct sanlk_lockspace ls;
+    PyObject *lockspace;
```
nirs: Changing the code should come after we update the tests. Doing changes before tests lead to broken tests that pass when the actual code fail.

aabawer:
i have parse args tests to be added, it will address the resource name and
lockspace names as xparams

tests/python_test.py
nirs: I posted this incomplete patch as example of how we should test this:
https://github.com/nirs/sanlock/commit/ff84992729c571db99b933d76226d1befec50013

@aabawer, what do you think?

aabawer:
- i think you should squash both commits for bytes notation for
resoruce and lockspace. in this commit i see only the bytes notation
for lockspace

- commit msg: "Need to update mark more python 3 tests as xfai, or
include the code" --- missing "l" in xfail

nirs:
@aabawer, this patch is incomplete example for the next version of your PR.

aabawer:
i have smaller tests for testing only the arg parsing for lockspace and
resoruce names.
maybe we dont want to increase the combinations of the existing tests..

aabawer:
i think that the last two patches for changing the parsing of lockspace and resource
should be squashed together for sake of tests operating in between those as we have tests checking 
the lockspace name and resource names with same xparam and we do not want to increase combinatorial complexity.

tests/python_test.py
```diff
@@ -487,3 +487,8 @@ def test_write_resource_invalid_disk(tmpdir, sanlock_daemon, disk):
         sanlock.write_resource("ls_name", "res_name", disks)
     assert repr(disk) in str(e.value)
 
+@pytest.mark.parametrize("filename,encoding", FILE_NAMES)
+def test_killpath(tmpdir, sanlock_daemon, filename, encoding):
+    cmd_path = util.generate_path(tmpdir, filename, encoding)
+    fd = sanlock.register()
+    sanlock.killpath(cmd_path, [], fd)
```
nirs: Testing empty arguments is nice, but I think it will be more useful if we add some arguments, maybe:

    sanlock.killpath(cmd_path, [cmd_path], fd)

tests/python_test.py
```diff
@@ -49,6 +51,21 @@
             reason="currently not supporting bytes paths")),
 ]
 
+LOCKSPACE_OR_RESOURCE_NAMES = [
+    # Bytes are supported with python 2 and 3.
+    pytest.param(
+        b"\xd7\x90",
+        marks=pytest.mark.xfail(six.PY3, reason="python 3 should support only bytes")),
```
nirs: The reason is more "not implemented yet".

tests/python_test.py
```diff
@@ -487,8 +504,85 @@ def test_write_resource_invalid_disk(tmpdir, sanlock_daemon, disk):
         sanlock.write_resource("ls_name", "res_name", disks)
     assert repr(disk) in str(e.value)
 
+
+@contextmanager
+def raises_sanlock_errno(expected_errno=errno.ECONNREFUSED):
```
nirs: This makes sense, but I remember different errors before - what did change?

tests/python_test.py
```diff
@@ -487,8 +504,85 @@ def test_write_resource_invalid_disk(tmpdir, sanlock_daemon, disk):
...
+@pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
+def test_add_lockspace_parse_args(name):
+    with raises_sanlock_errno():
+        sanlock.add_lockspace(name, 1, "ls_path", 0)
+
+
+@pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
+def test_write_lockspace_parse_args(name):
+    with raises_sanlock_errno():
+        sanlock.write_lockspace(name,"ls_path")
```
nirs: missing space after ,

tests/python_test.py
```diff
@@ -487,8 +504,85 @@ def test_write_resource_invalid_disk(tmpdir, sanlock_daemon, disk):
...
+@pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
+def test_reg_event_parse_args(name):
+    with raises_sanlock_errno():
+        sanlock.reg_event(name)
+
+@pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
+def test_end_event_parse_args(name):
+    with raises_sanlock_errno(errno.EALREADY):
+        sanlock.end_event(-1, name)
+
```
nirs: Strange that all this fail with a different error. Maybe David can explain this error?

The empty line at the end of the file can be removed.

tests/python_test.py
```diff
@@ -487,8 +504,85 @@ def test_write_resource_invalid_disk(tmpdir, sanlock_daemon, disk):
...
 def test_killpath(tmpdir, sanlock_daemon, filename, encoding):
     cmd_path = util.generate_path(tmpdir, filename, encoding)
     fd = sanlock.register()
     sanlock.killpath(cmd_path, [], fd)
+
+
+@pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
+def test_rem_lockspace_parse_args(name):
+    with raises_sanlock_errno():
+        sanlock.rem_lockspace(name, 1, "ls_path", 0)
```
nirs: "ls_path" and in other tests "disk_path" are fragile, because they can match a file in the current directory.
It would be more robust to use something like "/no/path/for/you". Since we use this value about 50 times, a constant would be useful.

*

tests/python_test.py
```diff
@@ -54,16 +54,16 @@
 LOCKSPACE_OR_RESOURCE_NAMES = [
     # Bytes are supported with python 2 and 3.
     pytest.param(
-        b"\xd7\x90",
-        marks=pytest.mark.xfail(six.PY3, reason="python 3 should support only bytes")),
+        b"\xd7\x90"),
+        #marks=pytest.mark.xfail(six.PY3, reason="python 3 should support only bytes")),
```
nirs: Why comment the marks? if this works now, we can remove the pytest.param() wrapper. This is not likely to change in the future.

tests/python_test.py
```diff
@@ -54,16 +54,16 @@
...
     # Python 2 also supports str.
     pytest.param(
         "\xd7\x90",
         marks=pytest.mark.skipif(six.PY3, reason="python 3 supports only bytes")),
     # Python 2 also supports unicode with ascii content.
     pytest.param(
         u"ascii",
-        marks=pytest.mark.skipif(six.PY3, reason="python 3 supports only bytes")),
+        marks=pytest.mark.xfail(reason="python 2 does not support unicode string, python 3 only supports bytes")),
```
nirs: This is a problem, since vdsm is using unicode names with ascii content now. This change will break vdsm using this code, or any other client that happen to use unicode strings. This is pretty common, since:

```
>>> import json
>>> json.loads(json.dumps({"name": "ascii"}))
{u'name': u'ascii'}
```

I think we will need to use "O&" format, and add a function that convert unicode
object to bytes on py2, and copy bytes object otherwise. Something like:

```
/* Return new reference via addr */
static int
convert_to_pybytes(PyObject *arg, void *addr)
{
#if PY_MAJOR_VERSION == 2
    if (PyUnicode_Check(arg)) {
        PyObject *bytes = PyUnicode_AsASCIIString(arg);
        if (bytes == NULL)
            return 0;

        *(PyObject **)addr = bytes;
        return 1;
    }
#endif

    if (PyBytes_Check(arg)) {
        Py_INCREF(arg);
        *(PyObject **)addr = arg;
        return 1;
    }
    
    set type error...
    return 0; 
}
```

nirs (changes requested):

tests/python_test.py
aabawer: if we expect bytes in Py3, unicode/ascii strings would still break there; why using bytes for lockspace/resource names anyway?
This brings back to me earlier comment that we could be happy with using ascii strings for those. py3 string would support it. maybe we just need to make sure we only allow ascii strings in py3 for lockspace and resoruce name for interoperability between py2 and py3 vdsms and document this limitation in the user-guide.

tests/python_test.py
aabawer: Probably we had another sanlock process still running from another crashed test.
so we had our foreground test communicating with the still active process and getting the ENOENT replies.
this can be easily reproduced:
1. run tox in one terminal for all tests
2. run tox -- -k parse_args in another terminal for the relevant tests
and you'll get the different errnos by mistake.

tests/python_test.py
aabawer: we can validate post-parsing or as a convertor, utilizing the following API:
`
PyObject* PyUnicode_AsASCIIString(PyObject *unicode)
    Return value: New reference.

    Encode a Unicode object using ASCII and return the result as Python bytes object. Error handling is “strict”. Return NULL if an exception was raised by the codec.
`

aabawer: we could use convert_to_pybytes shown above with same code (no ifdef) for both Py versions and not break anything on vdsm.

nirs: Supporting unicode and bytes is horrible hack, needed only for compatibillity with py 2. On py 3 code must use the correct type.

If we accept unicode, we should also return it in read_xxx, and we cannot do this because py 2 allowed anything, and there is no encoding.

aabawer: so this implies we'll need to straighten up the vdsm calls for lockspaces and resoruces once py3 sanlock is operational.

aabawer: anyway. i think the converter should be in a separate patch so it could be easily reverted once we drop py2 for good and then we can leave the "S" formatter for py3 as planned.

tests/python_test.py
aabawer: will apply dedicated patch  for dummy_path constant

tests/python_test.py
aabawer: see prev note about sanlock running in bg from another test.

python/sanlock.c
aabawer: code and test changed in same patch (squashed test here)

tests/util.py
```diff
@@ -160,3 +160,18 @@ def generate_path(dirname, filename, encoding=None):
     if encoding:
         path = path.encode(encoding)
     return path
+
+
+def sanlock_is_running():
+    exe_file = "./src/sanlock"
+    pid_file = "/tmp/sanlock/sanlock.pid"
```
nirs: /tmp/sanlock is define in tox, we should access it here via os.environ.

tests/util.py
```diff
@@ -160,3 +160,18 @@ def generate_path(dirname, filename, encoding=None):
     if encoding:
         path = path.encode(encoding)
     return path
+
+
+def sanlock_is_running():
+    exe_file = "./src/sanlock"
```
nirs: We can use SANLOCK here, or just use SANLOCK for comparing later.

tests/util.py
```diff
@@ -160,3 +160,18 @@ def generate_path(dirname, filename, encoding=None):
...
+    pid_file = "/tmp/sanlock/sanlock.pid"
+    try:
+        with open(pid_file, "r") as f:
+            for line in f:
```
nirs: pid file contain exactly one line, so we should do:

    pid = f.readline().strip()

tests/util.py
```diff
@@ -160,3 +160,18 @@ def generate_path(dirname, filename, encoding=None):
...
+                pid = line.strip()
+                if os.path.samefile(os.readlink("/proc/"+str(pid)+"/exe"), exe_file):
+                    return True
+    except OSError:
+        return False
+
+    return False
```
nirs: Would be little nicer and more correct as:

    proc_exec = "/proc/{}/exec".format(pid)

    try:
        executable = os.readlink(proc_exe)
        return os.path.samefile(executable, SANLOCK)
    except EnvironmentError as e:
        if e.errno != errno.ENOENT:
            raise

    return False

tests/python_test.py
```diff
@@ -487,8 +504,94 @@ def test_write_resource_invalid_disk(tmpdir, sanlock_daemon, disk):
...
+@contextmanager
+def raises_sanlock_errno(expected_errno=errno.ECONNREFUSED):
+    with pytest.raises(sanlock.SanlockException) as e:
+        yield
+    assert e.value.errno == expected_errno
+
+
+@pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
+def test_rem_lockspace_parse_args(name):
+    assert not util.sanlock_is_running(), "Test expects sanlock to be down"
```
nirs: This will cause the test to fail, instead of error.

Why not create a no_sanlock_daemon fixture, raising SanlockIsRunning exception
if util.sanlock_is_running() returns True?

    def test_rem_lockspace_parse_args(no_sanlock_daemon, name):
        ...

The current code duplicate the assert and the text for every test.

nirs (changes requested):
Reviewed up to:
python: Treat lockspace name argument as PyBytes

aabawer:
pushed PR updates.

python/sanlock.c
```diff
@@ -853,8 +853,8 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
 /* rem_lockspace */
 PyDoc_STRVAR(pydoc_rem_lockspace, "\
-rem_lockspace(lockspace, host_id, path, offset=0, async=False, unused=False)\n\
-Remove a lockspace, releasing the acquired host_id. If async is True the\n\
+rem_lockspace(lockspace, host_id, path, offset=0, wait=True, unused=False)\n\
+Remove a lockspace, releasing the acquired host_id. If wait is False the\n\
```
nirs: In python 2 we should have:

    rem_lockspace(lockspace, host_id, path, offset=0, async=False, unused=False, wait=True)

In pyhton 3:

    rem_lockspace(lockspace, host_id, path, offset=0, unused=False, wait=True)

I think the best way to do this is to create a tiny wrapper for python 2 and python 3, which will do only the parameter parsing, and then call the common code doing the operation with normalized arguments.

We can register either the correct version in sanlock_methods list using ifdef.

python/sanlock.c
```diff
@@ -864,30 +864,45 @@ successful termination these leases will be released.");
...
         convert_to_pybytes, &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset,
-        &async, &unused)) {
+        &wait, &unused)) {
+#if PY_MAJOR_VERSION == 2
+        /* Py2 fallback to check if we can parse with old async param instead */
+        int async = 0;
+        static char *legacy_kwlist[] = {"lockspace", "host_id", "path", "offset",
+                                            "async", "unused", NULL};
+        /* clear previous parse args errors */
+        PyErr_Clear();
```
nirs: We don't want to clear errors like this. We need 2 functions, each doing the right thing for the platform.

tests/python_test.py
```diff
@@ -587,3 +587,13 @@ def test_reg_event_parse_args(no_sanlock_daemon, name):
...
+
+@pytest.mark.parametrize("wait_key", [
+        pytest.param(
+            "async",
+            marks=pytest.mark.xfail(six.PY3,reason="deprecated key")),
+        pytest.param("wait")
+    ])
+def test_rem_lockspace_parse_args(no_sanlock_daemon, wait_key):
+    with raises_sanlock_errno():
+        sanlock.rem_lockspace(b"ls_name", 1, DUMMY_PATH, **{wait_key: False})
```
nirs: We need to test add_lockspace and remove lockspace with both async and wait on python 2. The best way would be to user parameters for the slow add_rem_lockspace and add_rem_lockspace_async, that will run both options on py2, and only wait= on py3.

nirs (changes requested):
Please do not add new changes to this PR. Every PR should be focus on one topic.

This work should be deferred until we finish name and path handling.

aabawer:
Okay about the later PR.

Regarding py2:

why would we like to allow:

rem_lockspace(lockspace, host_id, path, offset=0, async=False,
unused=False, wait=True)

it makes sense to call:

rem_lockspace(lockspace, host_id, path, offset=0, unused=False, wait=True)
or
rem_lockspace(lockspace, host_id, path, offset=0, async=False, unused=False)

but not simultaneously even if wait=True and async=False and they do not
conflict.

nirs:
We don't wan to allow call with both arguments, but the docstring should that both exist.

If this was written in pure python, we would have something like:

```
def call_py2(async=None, wait=None, *args, **kw):
    """
    Arguments:
        async (bool): Return before operation was done. deprecated, use wait instead.
            Will be removed in python 3.
        wait (bool): Wait until operation is done.
    """
    if async is not None and wait is not None:
        raise ...

   if wait is None:
       wait = True if async is None else not async
   
    _real_call(wait, *args, **kw)

def call_py3(wait=True, *args, **kw):
   """
    Arguments:
        wait (bool): Wait until operation is done.
   """
    _real_call(wait, *args, **kw)
```

This example is probably wrong but I it explains the idea.

aabawer:
Thanks for clarifying.

tests/conftest.py
```diff
@@ -44,3 +44,12 @@ def user_4k_path(request):
...
             "to enable 4k storage tests")
     return request.param
+
+
+@pytest.fixture
+def no_sanlock_daemon():
+    if util.sanlock_is_running():
+        raise SanlockIsRunning
+
+class SanlockIsRunning(BaseException):
```
nirs: Exceptions should always inherit from Exception. Inheriting from BaseException is needed very rarely.

tests/conftest.py
```diff
@@ -44,3 +44,12 @@ def user_4k_path(request):
             "user storage available - run 'python tests/strorage.py setup' "
             "to enable 4k storage tests")
     return request.param
+
+
+@pytest.fixture
+def no_sanlock_daemon():
+    if util.sanlock_is_running():
+        raise SanlockIsRunning
```
nirs: This works, but it would be nicer if the error expose the pid of the running process.

If we replace sanlock_is_running with:

    sanlock_status() -> {"pid": 1234, "running": True}

We can have a more helpful message here.

However I think this patch is good enough, so we can take it as is.

tests/util.py
```diff
@@ -160,3 +160,22 @@ def generate_path(dirname, filename, encoding=None):
...
+    pid_file = os.path.join(os.environ["SANLOCK_RUN_DIR"], "sanlock.pid")
+    try:
+        with open(pid_file, "r") as f:
+            pid = f.readline().strip()
+            executable = os.readlink("/proc/{}/exe".format(pid))
+            return os.path.samefile(executable, SANLOCK)
+    except OSError as e:
+        if e.errno != errno.ENOENT:
+            raise
+        return False
```
nirs: This return is not needed. If we do nothing, we will fall through to the next return.

tests/python_test.py
```diff
@@ -448,6 +448,7 @@ def test_write_resource_invalid_align_sector(
     path = str(tmpdir.join("resources"))
     util.create_file(path, MIN_RES_SIZE)
     disks = [(path, 0)]
+    util.sanlock_is_running()
```
nirs: Leftover?

python/sanlock.c
```diff
@@ -309,6 +309,47 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
     return Py_BuildValue("i", rv);
 }
 
+/*
+ * converts parsed arg into PyBytes object.
+ * if arg is not covertable as a Python2 Unicode or
+ * as a Bytes object we raise TypeError with a message
+ * and return 0. Otherwise we convert arg into PyBytes
+ * and store the new object reference in addr while
+ * returning 1.
```
nirs: I thin it would be better to explain what we actually do:

- if the object is a PyUnicode object, create a new PyBytes object with the content of arg encoded as ascii.
if encoding fail will raise xxx.
- If the object is already a PyBytes object, increase the reference count and return it via addr pointer.
- Otherwise raise ValueError and return 0.

python/sanlock.c
```diff
@@ -309,6 +309,47 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
...
+*/
+static int
+convert_to_pybytes(PyObject* arg, void *addr)
+{
+    if (arg == NULL)
+        return 1;
```
nirs: This is not possible, and we should crash if this happens.
We can do:

    assert(arg != NULL && "convert_to_pybytes called with NULL arg");

python/sanlock.c
```diff
@@ -309,6 +309,47 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
...
+        if (bytes == NULL)
+            return 0;
+
+        *(PyObject **)addr = bytes;
+        return 1;
+    }
+
+    if (!PyBytes_Check(arg)) {
+        set_value_error("Argument type is not bytes or unicode: %s", arg);
+        return 0;
```
nirs: Why duplicate the code for checking PyBytes?

The check on line 342 works for both cases:

- py2, and arg is not unicode object
- py3

python/sanlock.c
```diff
@@ -309,6 +309,47 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
...
+#endif
+
+    if (PyBytes_Check(arg)) {
+        Py_INCREF(arg);
+        *(PyObject **)addr = arg;
+        return 1;
+    }
+
+    set_value_error("Argument type is not bytes: %s", arg);
+    return 0;
```
nirs, line 346: We promised to raise a TypeError. We can make it easy by converting set_value_error to:

    set_error(PyObject *err, const *char fmt, PyObject *obj)

python/sanlock.c
```diff
@@ -347,10 +389,15 @@ py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock lockspace init failure");
-        return NULL;
+        goto exit_fail;
     }
 
+    Py_DECREF(lockspace);
```
nirs: Using exit_fail when we always need to cleanup duplicates the cleanup code, which lead to bugs when you forget to update both cleanup sections later.

I think it is better to use single return in these cases.

For example:

```
    int rv = -1;

    if (!parse_args(...))
        goto finally;

    rv = sanlock_direct_init(...);
    if (rv != 0)
        goto finally;

finally:
    Py_XDECREF(thing);
    if (rv != 0):
        return NULL;

    Py_RETURN_NONE;
```

python/sanlock.c
```diff
@@ -309,6 +309,47 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
...
+    }
+
+    set_value_error("Argument type is not bytes: %s", arg);
+    return 0;
+
+}
+
 /* init_lockspace */
 PyDoc_STRVAR(pydoc_init_lockspace, "\
 init_lockspace(lockspace, path, offset=0, max_hosts=0, num_hosts=0, \
```
nirs, line 351: We can remove this in python 3 - is is deprecated for at least 5.5 years. Vdsm also stopped
using it few month ago, but we cannot break it in python 2.

Lets add a TODO for removing it in python 3, or file an issue in pagure.

python/sanlock.c
```diff
@@ -398,11 +444,15 @@ py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
         goto exit_fail;
     }
 
+    Py_DECREF(lockspace);
+    Py_DECREF(resource);
```
nirs: Double cleanup again ...

nirs (changes requested):

python/sanlock.c
aabawer: i think its slightly over-detailed for a relatively small piece of code. i'd stick with the general function description.

nirs: The current version is misleading about converting PyBytes to PyBytes, lets make sure it is not.

aabawer:
Thanks for reviewing. updated the PR.

tests/python_test.py
aabawer: that would slow tests x 2  for py2

nirs: It will double the time for one or two tests, not all tests.

python/sanlock.c
```diff
@@ -347,10 +385,14 @@ py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
         __set_exception(rv, "Sanlock lockspace init failure");
-        return NULL;
+        goto finally;
     }
 
-    Py_RETURN_NONE;
+finally:
+    Py_XDECREF(lockspace);
+    if (rv == 0)
+        Py_RETURN_NONE;
```
nirs: I like to keep the happy path pattern - errors are check first, success is the normal path.

Keeping the same style everywhere helps to follow the code and avoid bugs.

python/sanlock.c
```diff
@@ -395,14 +436,15 @@ py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
 
-    free(res);
-    Py_RETURN_NONE;
-
-exit_fail:
-    free(res);
+finally:
+    Py_XDECREF(lockspace);
+    Py_XDECREF(resource);
+    if (res) free(res);
```
nirs, line 445: free(NULL) is ok, we don't need the if, but looks like this is the way free() is used in this module, so we can improve this later.

python/sanlock.c
```diff
@@ -737,9 +790,10 @@ acquired or released.");
 static PyObject *
 py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, waitrs = 0, flags = 0;
-    const char *lockspace, *path;
+    int rv = -1, waitrs = 0, flags = 0;
+    const char *path;
     struct sanlk_lockspace ls;
+    PyObject *lockspace = NULL;
```
nirs: I think the order of the local variables should match the usage:

- input arguments (e.g. lockspace)
- internal variables (e.g. struct sanlk_lockspace ls)
- return value (e.g. rv)

The current code does not keep this convention, but at least we can add the new lockpace where it was defined before.

Keeping this order also requires having one declaration per line, which is a good thing, making it harder to create too many variables.

Since this patch is large lets just make sure the new lockspace/resource variables are located near the place they were before the patch.

python/sanlock.c
```diff
@@ -943,6 +1004,7 @@ py_get_hosts(PyObject *self __unused, PyObject *args, PyObject *keywds)
     ls_list = __hosts_to_list(hss, hss_count);
 
 exit_fail:
```
nirs: For consistency, we should call this "finally" - this is not a failure path. Since this not an issue added by the patch we can fix it later.

python/sanlock.c
```diff
@@ -1527,29 +1603,33 @@ with -EBUSY error in this case.\n\
...
     if (rv < 0) {
         __set_exception(rv, "Unable to set event");
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv < 0)
+        return NULL;
```
nirs, line 1628: This is the pattern we need everywhere.

tests/python_test.py
```diff
@@ -54,16 +54,15 @@
...
+        b"\xd7\x90"),
     # Python 2 also supports str.
     pytest.param(
         "\xd7\x90",
         marks=pytest.mark.skipif(six.PY3, reason="python 3 supports only bytes")),
     # Python 2 also supports unicode with ascii content.
     pytest.param(
         u"ascii",
-        marks=pytest.mark.skipif(six.PY3, reason="python 3 supports only bytes")),
+        marks=pytest.mark.xfail(six.PY3, reason="python 3 only supports bytes")),
```
nirs, line 67: This is not expected failure, this input value is not valid on python 3, this is why we skip it. Lets keep the skip.

If you want to test behaviour with unicode names, you can add tests for raising TypeError, but I don't think it worth the effort. Maybe just one test to make sure convert_to_pybytes behaves correctly?

nirs (changes requested):
Thanks, merged all patches except the last 2 to my py3 branch. Please rebase.

tests/python_test.py
aabawer: IMO we expect py3 to have TypeError here for unicode input in current implementation. If it would XPASS for some reason it would hint us there is something wrong and undocumented with lockspace/resource parsing. If we skip it by default, we'll probably never find it.

aabawer:
Thanks, pushed fixes.

python/sanlock.c
```diff
@@ -149,7 +149,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
     *res_ret = res;
     return 0;
```
nirs, line 150: This function does not use try-finally pattern, but try-except. If we fail, we go to exit_fail and free res. If we succeed, we return res.

python/sanlock.c
```diff
@@ -235,18 +235,18 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
...
+            goto finally;
 
         if (PyList_Append(ls_list, ls_entry) != 0)
-            goto exit_fail;
+            goto finally;
 
         Py_DECREF(ls_entry);
     }
 
     return ls_list;
```
nirs, line 246: Same, we need to decref ls_list and ls_entry only on error, so this is not a finaly.

python/sanlock.c
```diff
@@ -592,17 +639,12 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
         "resource", rs->name,
         "version", rs->lver);
     if (rs_info  == NULL)
-        goto exit_fail;
+        goto finally;
 
-    /* success */
-    free(rs);
+finally:
+    if(rs) free(rs);
```
nirs, line 645: if() -> if ()

python/sanlock.c
```diff
@@ -592,17 +639,12 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
         "resource", rs->name,
         "version", rs->lver);
     if (rs_info  == NULL)
-        goto exit_fail;
+        goto finally;
```
nirs, line 642: We can remove this check and goto now.

python/sanlock.c
```diff
@@ -894,7 +951,7 @@ py_get_lockspaces(PyObject *self __unused, PyObject *args, PyObject *keywds)
     return ls_list;
```
nirs, line 951: This is not a finally pattern.

python/sanlock.c
```diff
@@ -1316,22 +1378,28 @@ unregister the event listener using end_event.");
...
+    fd = sanlock_reg_event(PyBytes_AsString(lockspace), NULL /* event */, 0 /* flags */);
     Py_END_ALLOW_THREADS
 
     if (fd < 0) {
         __set_exception(fd, "Unable to register event fd");
-        return NULL;
+        goto finally;
     }
 
     return Py_BuildValue("i", fd);
```
nirs, line 1396: This is not finally pattern.

python/sanlock.c
```diff
@@ -1316,22 +1378,28 @@ unregister the event listener using end_event.");
...
+
+finally:
+    Py_XDECREF(lockspace);
```
nirs, line 1399: Because this function does not go to finally in all cases, we leak the lockspace here.

python/sanlock.c
```diff
@@ -1376,70 +1444,70 @@ py_get_event(PyObject *self __unused, PyObject *args)
...
+            goto finally;
 
         Py_DECREF(item);
         item = NULL;
     }
 
     return events;
 
-exit_fail:
+finally:
```
nirs, line 1510: This is not finally pattern

tests/python_test.py
nirs: xfail is not mean for testing negative flows. It should be used to mark a test as expected failure until we have time to fix the code. Here there is nothing to fix, so we skip the test.

If you want to test that we raise TypeError, add one test calling with invalid type, and check that we raise the expected error.

This parameter will be removed once we drop python 2, but your negative test about type error should always run.

nirs (changes requested):
This patch is way to big and consume too much time to review.

I think this should be separated to these changes:
- change exit_fail used as finally to finally (but keep exit_fail used only on failures)
- add convert_to_pybytes function
- tests for new behavior, possibly marked as xfail when needed
- the implementation of the easy functions
- the more tricky functions, possibly one patch for each tricky function.

The trivial test refactoring adding constant is not related to this PR, please send it later when we finish with this work.

aabawer:
I agree that we dont have to make such a big patch.
In fact i think that we dont have to change the exit_fail/finally in most
occurrences as we usually have the following flow:
1. parse lockspace/resource/path
2. check for parse failure - here we should xdecref for safety
3. set parameters for sanlock api call - here we can decref any parsed
object after setup as we are done with those parsed objects
4. call sanock api

aabawer:
btw, we can use "__attribute__((cleanup))" for Internal PyObjects . its allowed since we already have __unused in code for "self" in api declarations, so the entire python/sanlock.c code is in pure gcc.
i think its a better option than using the "goto exit_fail" method.

nirs:
The issue is no change from exit_fail to finally, or using or not using gcc attributes, but trying to make the entire change at once in messy code.

For example having exit_fail that behave like finally, having multiple variables declarations per line, having double cleanup code. Fixing these issues before we add new code will make it much easier and safer to change the code

Getting rid of the temporary bytes object at the start of the function is require having cleanup twice, once for parsing, and then again after copying the values to sanlock structures. I don't want to have double cleanup code, this makes it harder to change the code.

Using more gcc attributes adds more complexity and risk, and may break with other compilers. We should use the same solution used by the upstream python source.

The solution suggested in the current code is good, we just need fix the incorrect changes, and split it to easy to review chunks. For example once the tests are correct, I can merge them and we don't have to 
continue reviewing them. But it the tests are part of the same huge commit I need to review them again each time.

aabawer:
okay

aabawer:
will deliver a fresh PR

PR #15, opened 2019-05-23
Py3 file paths

aabawer:
python: Apply pypath_converter(...) for file paths
    
In Python 3 we would like to handle paths as either unicode strings
or bytes. For this purpose we have a C-API called PyUnicode_FSConverter.
In this patch the missing API is implemented for Py2 based on Py3's [1]
and a unified no-cleanup behavior is enforced in Py3 to comply with Py2's usage.
    
[1] https://github.com/python/cpython/blob/master/Objects/unicodeobject.c#L3818
    
This PR should be merged after #14 is merged to nirs/py3

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -132,14 +132,14 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
 
         if (!PyTuple_Check(disk)) {
             set_error(PyExc_ValueError, "Invalid disk %s", disk);
-            goto exit_fail;
+            goto finally;
 
         }
 
         if (!PyArg_ParseTuple(disk, "sK", &path, &offset)) {
             /* Override the error since it confusing in this context. */
             set_error(PyExc_ValueError, "Invalid disk %s", disk);
-            goto exit_fail;
+            goto finally;
         }
 
         strncpy(res->disks[i].path, path, SANLK_PATH_LEN - 1);
@@ -149,7 +149,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
     *res_ret = res;
     return 0;
 
-exit_fail:
+finally:
     free(res);
     return -1;
 }
@@ -222,7 +222,7 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
 
     /* prepare the dictionary holding the information */
     if ((ls_list = PyList_New(0)) == NULL)
-        goto exit_fail;
+        goto finally;
 
     for (i = 0; i < hss_count; i++) {
 
@@ -235,10 +235,10 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
                 "io_timeout", hss[i].io_timeout,
                 "flags", hss[i].flags);
         if (ls_entry == NULL)
-            goto exit_fail;
+            goto finally;
 
         if (PyList_Append(ls_list, ls_entry) != 0)
-            goto exit_fail;
+            goto finally;
 
         Py_DECREF(ls_entry);
     }
@@ -246,7 +246,7 @@ __hosts_to_list(struct sanlk_host *hss, int hss_count)
     return ls_list;
 
     /* failure */
-exit_fail:
+finally:
     Py_XDECREF(ls_entry);
     Py_XDECREF(ls_list);
     return NULL;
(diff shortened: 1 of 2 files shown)


python/sanlock.c
```diff
@@ -71,6 +71,61 @@ __set_exception(int en, char *msg)
...
+    /* python 2 implementation */
+    PyObject *output = NULL;
+    Py_ssize_t size;
+    const char *data;
+
+    if (PyBytes_Check(arg)) {
+        Py_INCREF(arg);
+        output = arg;
+    }
+    else {
```
nirs: else should move to previous line:

```
} else {
```

python/sanlock.c
```diff
@@ -71,6 +71,61 @@ __set_exception(int en, char *msg)
...
+        output = PyUnicode_AsEncodedString(arg, Py_FileSystemDefaultEncoding, NULL);
+        if (!output)
+            return 0;
+        assert(PyBytes_Check(output));
+    }
+
+    size = PyBytes_GET_SIZE(output);
+    data = PyBytes_AS_STRING(output);
+    if ((size_t)size != strlen(data)) {
+        PyErr_Format(PyExc_ValueError, "Embedded null byte");
```
nirs, line 111: We can use set_error() so we see the repr() of the output in the error message, but we can also keep this as the python 3 version to make it easy to backport fixes to this code.

python/sanlock.c
```diff
@@ -124,8 +179,7 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
     res->num_disks = num_disks;
 
     for (i = 0; i < num_disks; i++) {
-        PyObject *disk;
-        const char *path;
+        PyObject *disk, *path = NULL;
```
nirs: Keeping one variable per line is best, and also a smaller patch.

python/sanlock.c
```diff
@@ -136,14 +190,16 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
 
         }
 
-        if (!PyArg_ParseTuple(disk, "sK", &path, &offset)) {
+        if (!PyArg_ParseTuple(disk, "O&K", pypath_converter, &path, &offset)) {
             /* Override the error since it confusing in this context. */
             set_error(PyExc_ValueError, "Invalid disk %s", disk);
+            Py_XDECREF(path);
```
nirs: Why cleanup here?

python/sanlock.c
```diff
@@ -136,14 +190,16 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
             goto exit_fail;
         }
 
-        strncpy(res->disks[i].path, path, SANLK_PATH_LEN - 1);
+        strncpy(res->disks[i].path, PyBytes_AsString(path), SANLK_PATH_LEN - 1);
         res->disks[i].offset = offset;
+        Py_DECREF(path);
```
nirs, line 199: And here?

And not add a finally label for cleanup?

nirs (changes requested):

python/sanlock.c
aabawer: the patch here is inside a for loop over resource disks (path, offset)
so we need to decref path object as soon as we can before next iteration, where we parse another path and loose the previous reference and cause a leak.

python/sanlock.c
aabawer: see prev comment.

tests/python_test.py
```diff
@@ -478,6 +478,7 @@ def test_write_resource_invalid_align_sector(
     path = str(tmpdir.join("resources"))
     util.create_file(path, MIN_RES_SIZE)
     disks = [(path, 0)]
+    util.sanlock_is_running()
```
nirs: Looks like leftover, I already merge this to my py3 branch:

commit f0f2a0ff50687cb06510582a845c970ab493c8f1
Author: Amit Bawer <abawer@redhat.com>
Date:   Tue May 21 14:35:40 2019 +0300

    tests: Add pytest.fixture of no_sanlock_daemon
    
    New fixture calls util.sanlock_is_running() to check if the
    pid showing in sanlock.pid file (if file exists) leads to local
    sanlock executable file to conclude if sanlock is currenlty
    running in test env. In case sanlock is running, a SunlockIsRunning
    exception will be raised from test.

python/sanlock.c
nirs: Make sense, but in this case path should be defined in the scope of the loop, and cleaned up inside the loop.

aabawer:
Oops... but if its not sent to David yet we can rebase -i and remove this
ineffective line.

nirs:
The patch I merge to my py3 branch does not include this line, you can just remove this patch from this PR.

aabawer:
OK

aabawer:
Replaced by #19

PR #16, opened 2019-05-26
Replace "async" with "wait" in {add,rem}_lockspace API

aabawer:
Since "async" became a reserved word in Python 3, we replace it
with the more commonly used term "wait" which serves the opposite
meaning, i.e. cause the operation to run async when wait=False.
    
For back compatibility we still support the old "async" flag in
Python 2. Using "async" and "wait" flags simultaneously would raise
RuntimeError in Python 2. For Python 3 API, "async" is deprecated.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -799,44 +799,99 @@ finally:
     Py_RETURN_NONE;
 }
 
+static int
+add_lockspace_parse_args(
+        PyObject *args,
+        PyObject *keywds,
+        struct sanlk_lockspace* ls,
+        int* flags,
+        uint32_t* iotimeout)
+{
+    int rv = 0;
+
+    *flags = 0;
+    *iotimeout = 0;
+    PyObject *lockspace = NULL, *path = NULL;
+    memset(ls, 0, sizeof(struct sanlk_lockspace));
+
+#if PY_MAJOR_VERSION == 2
+    int async = -1, wait = -1;
+    static char *kwlist[] = {"lockspace", "host_id", "path", "offset",
+                                "iotimeout", "async", "wait", NULL};
+
+    /* parse python 2 tuple */
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&kO&|kIii", kwlist,
+        convert_to_pybytes, &lockspace, &(ls->host_id), pypath_converter, &path, &(ls->host_id_disk.offset),
+        iotimeout, &async, &wait)) {
+        goto finally;
+    }
+    if (wait != -1 && async != -1) {
+        PyErr_Format(PyExc_RuntimeError, "async and wait flags must be mutually exclusive");
+        goto finally;
+    }
+    if (wait == -1) {
+        wait = 1;
+        if (async != -1)
+            wait = !async;
+    }
+#else
+    int wait = 1;
+    static char *kwlist[] = {"lockspace", "host_id", "path", "offset",
+                                "iotimeout","wait", NULL};
+
+    /* parse python 3 tuple */
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&kO&|kIi", kwlist,
+        convert_to_pybytes, &lockspace, &(ls->host_id), pypath_converter, &path, &(ls->host_id_disk.offset),
+        iotimeout, &wait)) {
+        goto finally;
+    }
+#endif
+    /* prepare sanlock_add_lockspace flags */
+    if (!wait) {
+        *flags |= SANLK_ADD_ASYNC;
+    }
+    /* prepare sanlock names */
+    strncpy(ls->name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
+    strncpy(ls->host_id_disk.path, PyBytes_AsString(path), SANLK_PATH_LEN - 1);
+    rv = 1;
+
+finally:
+    Py_XDECREF(lockspace);
+    Py_XDECREF(path);
+    return rv;
+}
+
 /* add_lockspace */
+#if PY_MAJOR_VERSION == 2
 PyDoc_STRVAR(pydoc_add_lockspace, "\
-add_lockspace(lockspace, host_id, path, offset=0, iotimeout=0, async=False)\n\
-Add a lockspace, acquiring a host_id in it. If async is True the function\n\
+add_lockspace(lockspace, host_id, path, offset=0, iotimeout=0, async=False, wait=True)\n\
+Add a lockspace, acquiring a host_id in it. If wait is False the function\n\
+will return immediatly and the status can be checked using inq_lockspace;\n\
+This behavior is identical to having the async flag set as True. async flag\n\
+is maintained for back compatibility with earlier use cases and will be\n\
+deprecated in sanlock release for python 3.\n\
+The iotimeout option configures the io timeout for the specific lockspace,\n\
+overriding the default value (see the sanlock daemon parameter -o).");
+#else
+PyDoc_STRVAR(pydoc_add_lockspace, "\
+add_lockspace(lockspace, host_id, path, offset=0, iotimeout=0, wait=True)\n\
+Add a lockspace, acquiring a host_id in it. If wait is False the function\n\
 will return immediatly and the status can be checked using inq_lockspace.\n\
 The iotimeout option configures the io timeout for the specific lockspace,\n\
 overriding the default value (see the sanlock daemon parameter -o).");
+#endif
 
 static PyObject *
 py_add_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv = -1, async = 0, flags = 0;
+    int rv = 0, flags = 0;
     uint32_t iotimeout = 0;
-    PyObject *lockspace = NULL;
-    PyObject *path = NULL;
     struct sanlk_lockspace ls;
 
-    static char *kwlist[] = {"lockspace", "host_id", "path", "offset",
-                                "iotimeout", "async", NULL};
-
-    /* initialize lockspace structure */
-    memset(&ls, 0, sizeof(struct sanlk_lockspace));
-
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&kO&|kIi", kwlist,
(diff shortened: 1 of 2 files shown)


python/sanlock.c
```diff
@@ -536,9 +536,8 @@ py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     int rv = -1, max_hosts = 0, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
     uint32_t io_timeout = 0;
-    PyObject *lockspace = NULL;
-    PyObject *path = NULL;
     struct sanlk_lockspace ls;
+    PyObject *lockspace = NULL, *path = NULL;
```
nirs: This change is not needed and unwanted. I like to have one argument per line. This makes the code easier to follow and modify (creating smaller and more clear diffs).

python/sanlock.c
```diff
@@ -595,9 +594,8 @@ py_read_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     int rv = -1, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
     uint32_t io_timeout = 0;
-    PyObject *path = NULL;
     struct sanlk_lockspace ls;
-    PyObject *ls_info = NULL;
+    PyObject *ls_info = NULL, *path = NULL;
```
nirs: Please keep previous code, change not needed.

python/sanlock.c
```diff
@@ -661,9 +659,8 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
     int rv = -1, rs_len, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
-    PyObject *path = NULL;
     struct sanlk_resource *rs;
-    PyObject *rs_info = NULL;
+    PyObject *rs_info = NULL, *path = NULL;
```
nirs: Not needed

tests/python_test.py
```diff
@@ -332,8 +332,12 @@ def test_add_rem_lockspace(tmpdir, sanlock_daemon, size, offset):
     lockspaces = sanlock.get_lockspaces()
     assert lockspaces == []
 
-
-def test_add_rem_lockspace_async(tmpdir, sanlock_daemon):
+@pytest.mark.parametrize("flags", [
+        pytest.param({"async": True},
+            marks=pytest.mark.skipif(six.PY3, reason="async flag is deprecated by wait")),
```
nirs, line 336: "is depracated by" -> "was replaced by"

tests/python_test.py
```diff
@@ -512,10 +516,21 @@ def raises_sanlock_errno(expected_errno=errno.ECONNREFUSED):
 
 @pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
 @pytest.mark.parametrize("filename,encoding", FILE_NAMES)
-def test_rem_lockspace_parse_args(no_sanlock_daemon, name, filename, encoding):
+@pytest.mark.parametrize("flags", [
+        pytest.param({"wait": True, "async": False},
+            marks=[
+                    pytest.mark.xfail(six.PY2, raises=RuntimeError, reason="async and wait flags must be mutually exclusive"),
+                    pytest.mark.xfail(six.PY3, raises=TypeError, reason="async flag is deprecated by wait"),
+                ]),
```
nirs: Both cases should have negative test:

- using async= on py3 raises TypeError
- using both wait= and async= on python2 raises RuntimeError

We don't use xfail as a way to test negative flows. xfail is a way to keep the tests pass before we can fix the real issue.

python/sanlock.c
```diff
@@ -918,50 +915,103 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+
+    /* prepare sanlock_rem_lockspace flags */
+    if (!wait) {
+        *flags |= SANLK_REM_ASYNC;
+    }
+
+    if (unused) {
+        *flags |= SANLK_REM_UNUSED;
+    }
+    rv = 1;
```
nirs, line 1028: Can we have smaller and simpler functions that only do the argument parsing, and call common
function to do the actual action?

For example:

```
def rem_lockspace_py2(async=None, wait=None):
    parse args for py2
    rem_lockspace_internal(wait)

def rem_lockspace_py3(wait):
    parse args for py3
    rem_lockspace_internal(wait)
```
We can register the the py2 version when building py2 extension, and the py3 version when building py3 version.

python/sanlock.c
```diff
@@ -918,50 +915,103 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+#else
+PyDoc_STRVAR(pydoc_rem_lockspace, "\
+rem_lockspace(lockspace, host_id, path, offset=0, unused=False, wait=True)\n\
+Remove a lockspace, releasing the acquired host_id. If wait is False the\n\
 function will return immediately and the status can be checked using\n\
 inq_lockspace. If unused is True the command will fail (EBUSY) if there is\n\
 at least one acquired resource in the lockspace. Otherwise (the default)\n\
 sanlock will try to terminate processes holding resource leases and upon\n\
 successful termination these leases will be released.");
+#endif
```
nirs, line 1058: Having the ifdef around the line registering the function and doc would be nicer.

python/sanlock.c
```diff
@@ -918,50 +915,103 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-    strncpy(ls.host_id_disk.path, PyBytes_AsString(path), SANLK_PATH_LEN - 1);
-
-    /* prepare sanlock_rem_lockspace flags */
-    if (async) {
-        flags |= SANLK_REM_ASYNC;
-    }
-
-    if (unused) {
-        flags |= SANLK_REM_UNUSED;
-    }
```
nirs, line 964: This block can be shared by both py2 and py3 versions.

python/sanlock.c
```diff
@@ -918,50 +915,103 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-    /* initialize lockspace structure */
-    memset(&ls, 0, sizeof(struct sanlk_lockspace));
-
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&kO&|kii", kwlist,
-        convert_to_pybytes, &lockspace, &ls.host_id, pypath_converter, &path,
-        &ls.host_id_disk.offset,
-        &async, &unused)) {
-        goto finally;
-    }
```
nirs, line 951: Only the code above is relevant for argument parsing and handling wait/async.

nirs (changes requested):
As usual, I think we need to start with the tests. Can you extract them to the first commit?

nirs:
Lets wait with this work until we get 3.8.0 out. I feel we are spending too much time on python 2 compatibility, while we can hack around this difference in vdsm easily:

```
    if six.PY3:
        extra = {"wait": False}
    else:
        extra = {"async": True}
```

We will have to do something like this in vdsm anyway until we can drop python 2 support in vdsm, which is few month a head.

aabawer:
I saw the alternative patch you delivered, this means there is no "async"
flag anyway once we work with sanlock 3.8.0, so i guess the fallback
mentioned for vdsm is just to be on the safe side while still working with
older sanlock in py2.
But in case we could use sanlock 3.8.0 with py2, I'd suggest to fallback
regardless of python version: that is, if got TypeError from
sanlock.{add,rem}_lockspace  with "wait" retry with "async" (with the
opposite value).
Probably no need for further changes in sanlock on this issue since only
vdsm is using it, and if someone else would start, he'll only know about
the "wait" flag.

nirs:
The compatibility mechanism in vdsm is required if we need to consume sanlock 3.8 on python 2.
On python 3 we never planned to support async=, so we need a way to change the behavior
in vdsm anyway.

If we never ship sanlock 3.8 on python 2 (e.g. Fedora 29), we will need the compatibility code only when we start using sanlock 3.8 on python 3 (e.g. Fedora 30). I think this is a valid option since we will not have vdsm for python 3 on Fedora 29 anyway, we plan to build it only for Fedora 30.

aabawer:
no?


    if six.PY3:
        extra = {"wait": False}
    else:
        extra = {"async": True}

        try:

        sanlock.add_lockspace(...,extra)

    except TypeError:

         if six.PY2:

            sanlock.add_lockspace(..., {"wait": False})     # if we
use sanlock 3.8.0 with py2

nirs:
On python 3 we know that wait is used, on python 2 we know that async is used, so there is no reason to try and handle TypeError, which may be caused by another issues.

python/sanlock.c
aabawer: we can, but it will be ugly just the same as will need to duplicate the text for py3_doc and py2_doc strings and then use the ifdef in the api listing array instead.

python/sanlock.c
aabawer: i like that rem_lockspace_parse_args() hides the cleanup from its caller, thanks to that the caller doesn't have to deal with anything not related the sanlock API call itself

python/sanlock.c
aabawer: it is shared in rem_lockspace_parse_args()

tests/python_test.py
aabawer: think it has the same meaning. but will change

python/sanlock.c
aabawer: i think its more readable to have the ifdef in the place the functionality is added for both version instead having to duplicate the methods and follow their registration to figure which one is actually used.

python/sanlock.c
aabawer: we can, but it will amount to same amount of code. and i dont see any difference.

aabawer:
updated the patch just in case someone will need sanlock3.8.1 with py2 on fc29

aabawer:
> As usual, I think we need to start with the tests. Can you extract them to the first commit?

in this case we already have async flag tests in the test set, "wait" has never existed in the past time for this api so i see no particular reason to add extra tests before the patch, just to mark them all as xfails.

nirs:
This is not relevant now, closing.

PR #17, opened 2019-05-26
tests: Remove stale symlinks to loop device on storage setup 

aabawer:
os.symlink(device, link_path) may fail - so we need to delete stale symlink in setup.

This will happen since loop devices create for tests are not persistent between reboots,
so doing:

    python tests/storage.py setup
    reboot
    python tests/storage.py setup

Will result in stale symlinks for non-existing loop devices on last call
to storage setup.

### DIFF
--- a/tests/storage.py
+++ b/tests/storage.py
@@ -84,6 +84,9 @@ def create_loop_device(link_path, backing_file, size=1024**3,
 
     device = out.decode("utf-8").strip()
 
+    #remove stale symlink
+    if os.path.islink(link_path):
+        os.unlink(link_path)
     os.symlink(device, link_path)
     chown(link_path)
 


nirs:
> Since symlinks are created under /var/tmp/ which is eventually cleared
> by the system, 

The system never clears /var/tmp.

> we may get to a point where the setup will fail for
> an already existing loop device but missing symlinks 

If we have a loop device without a symlink, it is not our loop device, so there is no problem.
Calling teardown will not do anything useful.

> and vice-versa.

If we have symlink without devices, this check in setup will fail:

 46     if not os.path.exists(LOOP1):                                                                                                                                        
 47         create_loop_device(LOOP1, BACKING1)

And setup will create a new loop device (good).

os.symlink(device, link_path) may fail - so we may need to delete stale symlink in setup.

I think this always happen now if you do:

1. python tests/storage.py setup
2. reboot
3. python tests/storage.py setup

> Adding a teardown before setup will assure we can take all required
> steps to make tests 4k ready.

This fix is not correct. The result is re-creating perfectly good setup for no reason, and having to enter the password again.

The correct fix is to delete stale symlink.

aabawer:
fixed the PR. thanks

tests/storage.py
```diff
@@ -138,5 +140,15 @@ def remove_file(path):
...
 
 
+def remove_stale_symlink(path):
+    try:
+        if not os.path.exists(os.readlink(path)):
+            log.info("Removing stale symlink %s", path)
+            os.unlink(path)
+    except EnvironmentError as e:
+        if e.errno != errno.ENOENT:
+            raise
```
nirs: I think we can simplify this a lot by adding this wrapper for creating symlink:

```
def create_symlink(src, dst):
    if os.path.islink(dst):
        os.unlink(dst)
    os.symlink(src, dst)
```

nirs (changes requested):

aabawer:
we have a precondition for all the loop device linking and making at line
48:
if not os.path.exists(LOOP1):
where LOOP1 is the link path, so its better to clean the dangling links
beforehand.

nirs:
> we have a precondition for all the loop device linking and making at line 48: if not os.path.exists(LOOP1): where LOOP1 is the link path, so its better to clean the dangling links beforehand.

os.path.exist() check if the link target exist. When it return False we know that there is no loop device. We have two options:
- there is a stale link
- there is no link

os.path.islink() detect if we have a link that should be deleted.

Here is example flow:
```
>>> import os
>>> os.symlink("/no/such/path", "link")
>>> os.path.exists("link")
False
>>> os.path.islink("link")
True
>>> os.unlink("link")
>>> os.path.exists("link")
False
>>> os.path.islink("link")
False
```

aabawer:
Nice. Fixed the PR

nirs (approved):
Thanks!

PR #18, opened 2019-06-02
Parse lockspace and resource names as bytes

aabawer:
PR outline:
1. set hardcoded ls_names and res_names as bytes in tests.
2. shut tests involving lockspace or resource names as bytes as xfail for py3.
3. add pybytes converter to sanlock.c 
4. go over API and apply parser per each lockspace/resource API call in a dedicated patch for easy review.
5. re-enable xfail tests for py3.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -854,8 +854,9 @@ successful termination these leases will be released.");
 static PyObject *
 py_rem_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, async = 0, unused = 0, flags = 0;
-    const char *lockspace, *path;
+    int rv = -1, async = 0, unused = 0, flags = 0;
+    PyObject *lockspace = NULL;
+    const char *path;
     struct sanlk_lockspace ls;
 
     static char *kwlist[] = {"lockspace", "host_id", "path", "offset",
@@ -865,14 +866,14 @@ py_rem_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "sks|kii", kwlist,
-        &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset, &async,
-        &unused)) {
-        return NULL;
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&ks|kii", kwlist,
+        convert_to_pybytes, &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset,
+        &async, &unused)) {
+        goto finally;
     }
 
     /* prepare sanlock names */
-    strncpy(ls.name, lockspace, SANLK_NAME_LEN);
+    strncpy(ls.name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(ls.host_id_disk.path, path, SANLK_PATH_LEN - 1);
 
     /* prepare sanlock_rem_lockspace flags */
@@ -891,9 +892,13 @@ py_rem_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock lockspace remove failure");
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
 }
 
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -1538,14 +1538,15 @@ static PyObject *
 py_end_event(PyObject *self __unused, PyObject *args)
 {
     int fd = -1;
-    const char *lockspace = NULL;
-    int rv;
+    PyObject *lockspace = NULL;
+    int rv = -1;
 
-    if (!PyArg_ParseTuple(args, "is", &fd, &lockspace))
-        return NULL;
+    if (!PyArg_ParseTuple(args, "iO&", &fd, convert_to_pybytes, &lockspace)) {
+        goto finally;
+    }
 
     Py_BEGIN_ALLOW_THREADS
-    rv = sanlock_end_event(fd, lockspace, 0 /* flags */);
+    rv = sanlock_end_event(fd, PyBytes_AsString(lockspace), 0 /* flags */);
     Py_END_ALLOW_THREADS
 
     if (rv < 0) {
@@ -1553,6 +1554,10 @@ py_end_event(PyObject *self __unused, PyObject *args)
         return NULL;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv < 0)
+        return NULL;
     Py_RETURN_NONE;
 }
 
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -32,6 +32,8 @@
 
 #define MODULE_NAME "sanlock"
 
+#define BIND_ERROR -1000
+
 /* Functions prototypes */
 static void __set_exception(int en, char *msg) __sets_exception;
 static int __parse_resource(PyObject *obj, struct sanlk_resource **res_ret) __neg_sets_exception;
@@ -791,8 +793,9 @@ acquired or released.");
 static PyObject *
 py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, waitrs = 0, flags = 0;
-    const char *lockspace, *path;
+    int rv = BIND_ERROR, waitrs = 0, flags = 0;
+    PyObject *lockspace = NULL;
+    const char *path;
     struct sanlk_lockspace ls;
 
     static char *kwlist[] = {"lockspace", "host_id", "path", "offset",
@@ -802,10 +805,10 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "sks|ki", kwlist,
-        &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&ks|ki", kwlist,
+        convert_to_pybytes, &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset,
         &waitrs)) {
-        return NULL;
+        goto finally;
     }
 
     /* prepare sanlock_inq_lockspace flags */
@@ -814,7 +817,7 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     }
 
     /* prepare sanlock names */
-    strncpy(ls.name, lockspace, SANLK_NAME_LEN);
+    strncpy(ls.name, PyBytes_AsString(lockspace), SANLK_NAME_LEN);
     strncpy(ls.host_id_disk.path, path, SANLK_PATH_LEN - 1);
 
     /* add sanlock lockspace (gil disabled) */
@@ -822,13 +825,17 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     rv = sanlock_inq_lockspace(&ls, flags);
     Py_END_ALLOW_THREADS
 
-    if (rv == 0) {
+finally:
+    Py_XDECREF(lockspace);
+
+    if (rv == BIND_ERROR)
+        return NULL;
+    if (rv == 0)
         Py_RETURN_TRUE;
-    } else if (rv == -ENOENT) {
+    if (rv == -ENOENT)
         Py_RETURN_FALSE;
-    } else if (rv == -EINPROGRESS) {
+    if (rv == -EINPROGRESS)
         Py_RETURN_NONE;
-    }
 
     __set_exception(rv, "Sanlock lockspace inquire failure");
     return NULL;
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -460,10 +460,11 @@ Sector can be one of (512, 4096).");
 static PyObject *
 py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, max_hosts = 0, sector = SECTOR_SIZE_512;
+    int rv = -1, max_hosts = 0, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
     uint32_t io_timeout = 0;
-    const char *lockspace, *path;
+    PyObject *lockspace = NULL;
+    const char  *path;
     struct sanlk_lockspace ls;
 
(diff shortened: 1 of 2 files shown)


tests/python_test.py
```diff
@@ -66,6 +66,7 @@
         marks=pytest.mark.skipif(six.PY3, reason="python 3 supports only bytes")),
```
nirs: Makes sense. With these change we covered all inputs on python 2 and 3.

python/sanlock.c
```diff
@@ -309,6 +309,43 @@ py_get_alignment(PyObject *self __unused, PyObject *args)
     return Py_BuildValue("i", rv);
```
nirs, line 311: Looks good.

python/sanlock.c
```diff
@@ -384,9 +385,13 @@ py_init_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
 }
```
nirs, line 399: Looks good, but the parse_args test can be fixed now.

python/sanlock.c
```diff
@@ -437,15 +437,16 @@ py_init_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
-
-exit_fail:
-    free(res);
-    return NULL;
 }
 
 /* write_lockspace */
```
nirs, line 454: Looks good but the parse_args tests can be fixed now.

python/sanlock.c
```diff
@@ -497,9 +498,13 @@ py_write_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
 }
```
nirs, line 512: With this test can should be able to make test_write_lockspace_parse_args pass without xfail. This is the advantage of incremental development.

python/sanlock.c
```diff
@@ -711,15 +711,16 @@ py_write_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
-
-exit_fail:
-    free(rs);
-    return NULL;
 }
 
 /* add_lockspace */
```
nirs, line 728: We should be able to make test_write_resource_parse_args pass without xfail now.

python/sanlock.c
```diff
@@ -768,9 +769,13 @@ py_add_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv != 0 )
+        return NULL;
     Py_RETURN_NONE;
 }
```
nirs, line 783: test_add_lockspace_parse_args should pass now without xfail.

python/sanlock.c
```diff
@@ -802,10 +803,10 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
     memset(&ls, 0, sizeof(struct sanlk_lockspace));
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "sks|ki", kwlist,
-        &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&ks|ki", kwlist,
+        convert_to_pybytes, &lockspace, &ls.host_id, &path, &ls.host_id_disk.offset,
         &waitrs)) {
-        return NULL;
+        goto finally;
```
nirs, line 811: If we fail here, we already set the exception...

python/sanlock.c
```diff
@@ -814,14 +815,17 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
     Py_BEGIN_ALLOW_THREADS
     rv = sanlock_inq_lockspace(&ls, flags);
     Py_END_ALLOW_THREADS
 
+finally:
+    Py_XDECREF(lockspace);
+
     if (rv == 0) {
         Py_RETURN_TRUE;
     } else if (rv == -ENOENT) {
```
nirs: But the code bellow this (hidden by github) is:

```
 771     if (rv == 0) {
 772         Py_RETURN_TRUE;
 773     } else if (rv == -ENOENT) {
 774         Py_RETURN_FALSE;
 775     } else if (rv == -EINPROGRESS) {
 776         Py_RETURN_NONE;
 777     }
 778 
 779     __set_exception(rv, "Sanlock lockspace inquire failure");
 780     return NULL;
 781 }
 ```

So we will set bogus sanlock excpetion with rv=-1, which will be translated to 
errno 1 (EPERM) in __set_exception(), so on the python side we will see:

    sanlock.SanlockException: (1, 'Sanlock lockspace inquire failure', 'Operation not permitted')

One way to avoid it is to use a sentinel value that cannot be returned by sanlock. sanlock returns either system errors (-1 - -128) or sanlock errors (-201 - -278) (src/sanlock_rv.h).

So we can do:

```
#define BINDING_ERROR -1000
...
int rv = BINDING_ERROR;
...
    if (rv == 0) {
        Py_RETURN_TRUE;
    } else if (rv == -ENOENT) {
        Py_RETURN_FALSE;
    } else if (rv == -EINPROGRESS) {
        Py_RETURN_NONE;
    } else if (rv == BINDING_ERROR) {
        return NULL;
    }

    __set_exception(rv, "Sanlock lockspace inquire failure");
    return NULL;
}
```

File: python/sanlock.c [hunk 10]
```diff
@@ -888,9 +889,13 @@ py_rem_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
-        return NULL;
+        goto finally;
     }
 
+finally:
+    Py_XDECREF(lockspace);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
 }
```
nirs, line 904: test_rem_lockspace_parse_args should pass without xfail with this change.

python/sanlock.c
```diff
@@ -979,34 +979,37 @@ timestamp and the io_timeout.\n");
...
 
-exit_fail:
-    if (hss) free(hss);
+finally:
+    Py_XDECREF(lockspace);
+    free(hss);
+    if (rv < 0)
+        return NULL;
     return ls_list;
 }
```
nirs, line 1018: Looks good, but the relevant test should be modified.

python/sanlock.c
```diff
@@ -1078,15 +1078,16 @@ py_acquire(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+    Py_XDECREF(resource);
     free(res);
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
-
-exit_fail:
-    free(res);
-    return NULL;
 }
```
nirs, line 1095: Looks good, relevant tests should be modified.

python/sanlock.c
```diff
@@ -1129,15 +1130,16 @@ py_release(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
 
 /* request */
```
nirs, line 1148: Looks good, missing test change.

python/sanlock.c
```diff
@@ -1197,15 +1198,16 @@ py_request(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+    if (rv != 0)
+        return NULL;
     Py_RETURN_NONE;
-
-exit_fail:
-    free(res);
-    return NULL;
 }
 
 /* read_resource_owners */
```
nirs, line 1216: Looks good, missing test change.

python/sanlock.c
```diff
@@ -1265,14 +1266,18 @@ py_read_resource_owners(PyObject *self __unused, PyObject *args, PyObject *keywd
...
-    if (hss) free(hss);
+finally:
+    Py_XDECREF(lockspace);
+    Py_XDECREF(resource);
+    free(res);
+    free(hss);
+    if (rv != 0)
+        return NULL;
     return ls_list;
 }
```
nirs, line 1286: Looks good, missing test change.

python/sanlock.c
```diff
@@ -1392,16 +1392,20 @@ unregister the event listener using end_event.");
...
+        Py_XDECREF(lockspace);
         return NULL;
+    }
 
     Py_BEGIN_ALLOW_THREADS
-    fd = sanlock_reg_event(lockspace, NULL /* event */, 0 /* flags */);
+    fd = sanlock_reg_event(PyBytes_AsString(lockspace), NULL /* event */, 0 /* flags */);
     Py_END_ALLOW_THREADS
 
+    Py_DECREF(lockspace);
```
nirs: Why not add proper cleanup here like in previous patches?

We can do:

```
    if (fd < 0) {
        __set_exception(fd, "Unable to register event fd");
       goto finally;
    }

finally:
    Py_XDECREF(lockspace);

    if (fd < 0)
        return NULL;

    return Py_BuildValue("i", fd);
```

python/sanlock.c
```diff
@@ -1534,16 +1534,20 @@ static PyObject *
...
     Py_BEGIN_ALLOW_THREADS
-    rv = sanlock_end_event(fd, lockspace, 0 /* flags */);
+    rv = sanlock_end_event(fd, PyBytes_AsString(lockspace), 0 /* flags */);
     Py_END_ALLOW_THREADS
 
+    Py_DECREF(lockspace);
+
     if (rv < 0) {
         __set_exception(rv, "Unable to unregister event fd");
         return NULL;
```
nirs, line 1554: Should use same finally pattern like previous patch.

python/sanlock.c
```diff
@@ -1611,24 +1611,27 @@ with -EBUSY error in this case.\n\
...
     Py_BEGIN_ALLOW_THREADS
-    rv = sanlock_set_event(lockspace, &he, flags);
+    rv = sanlock_set_event(PyBytes_AsString(lockspace), &he, flags);
     Py_END_ALLOW_THREADS
 
+    Py_DECREF(lockspace);
+
     if (rv < 0) {
         __set_exception(rv, "Unable to set event");
         return NULL;
```
nirs: Should use same finally pattern as previous patch.

tests/python_test.py
```diff
@@ -53,9 +53,7 @@
 
 LOCKSPACE_OR_RESOURCE_NAMES = [
     # Bytes are supported with python 2 and 3.
-    pytest.param(
-        b"\xd7\x90",
-        marks=pytest.mark.xfail(six.PY3, reason="bytes support not implemented yet")),
+    pytest.param(b"\xd7\x90"),
```
nirs, line 56: We don't need the pyparam() wrapper now, but we can keep it for symmetry if you like.

nirs (changes requested):
I can take some of the first patches now.

Most patches are ready, but it would be nice to fix the xfail for parse_args tests related to each patch
to prove that the patch works. With this I can merge most patches. If you cannot find an easy way to enable the tests relevant to the patches, I can take all the patches together.

Some patches are missing proper cleanup, adding it should be easy and I don't see reason to skip it and have double cleanup code.

nirs:
Here is a simple way to enable the relevant tests for every patch:
https://github.com/nirs/sanlock/commits/amit-py3-incremental-tests
- https://github.com/nirs/sanlock/commit/b26ebb3c3337a51a3e3df17537b44dda899cc85a
  (this code was missing a test - best add the tests before as separate patch)
- https://github.com/nirs/sanlock/commit/21285c8edc3d813f926bbad819593e1710c3ceb9
  (same, added missing test)
- https://github.com/nirs/sanlock/commit/1e2af7e708224ad151c16963c5e9a6f197727be1
  (1 line change to make the test pass)

What do you think?

python/sanlock.c
aabawer: We can. But it will larger than actual cleanup to be done in this case. It will be only for lockspace. We can always add finaly label when time will come to add more cleanups.

python/sanlock.c
aabawer: See prev comment

python/sanlock.c
aabawer: See prev comment

python/sanlock.c
nirs: Not sure what is "larger then actual cleanup". Why not keep the code consistent?

tests/python_test.py
aabawer: i know

aabawer:
1.will add the cleanup for the small ones as well.
2.will add parse args tests for init_{lockspace, resource} as part of their API commits 
3.i think this PR has matured to the level it can be merged entirely after 1 & 2 so i dont think we should waste time on enabling each test that can be enabled per every patch in the PR, but leave it to the final test patch as it is.

aabawer:
Updated PR with fixes.

nirs:
I think we can take this PR in the current form, but the next PRs should have working test with the code implementing the change.

aabawer:
good news.
will try, some tests involve more than a single API and i'd like to change
one API at a time for sake of review.

nirs:
The big flow tests should not be affected as since the default path can be unicode with ascii content, which is already supported.  We can add small tests like the tests for the lockspace/resource names to make it easy to fix the API in small steps.

python/sanlock.c
```diff
@@ -814,21 +817,25 @@ py_inq_lockspace(PyObject *self __unused, PyObject *args, PyObject *keywds)
...
+        return NULL;
+    if (rv == 0)
         Py_RETURN_TRUE;
-    } else if (rv == -ENOENT) {
+    if (rv == -ENOENT)
         Py_RETURN_FALSE;
-    } else if (rv == -EINPROGRESS) {
+    if (rv == -EINPROGRESS)
         Py_RETURN_NONE;
-    }
```
nirs, line 777: The bracing changes are not needed. I'll fix this locally before pushing.

nirs (approved):

tests/python_test.py
```diff
@@ -603,3 +603,8 @@ def test_reg_event_parse_args(no_sanlock_daemon, name):
 def test_end_event_parse_args(no_sanlock_daemon, name):
     with raises_sanlock_errno(errno.EALREADY):
         sanlock.end_event(-1, name)
+
```
nirs, line 587: Missing empty line, I will fix it locally before I push.

python/sanlock.c
```diff
@@ -1538,21 +1538,26 @@ static PyObject *
...
+    }
 
     Py_BEGIN_ALLOW_THREADS
-    rv = sanlock_end_event(fd, lockspace, 0 /* flags */);
+    rv = sanlock_end_event(fd, PyBytes_AsString(lockspace), 0 /* flags */);
     Py_END_ALLOW_THREADS
 
     if (rv < 0) {
         __set_exception(rv, "Unable to unregister event fd");
         return NULL;
```
nirs, line 1554: There is a leak here. Will fix locally before pushing.

nirs:
I merge this into my py3 branch as 1ef37674d0465c4cf37c0892334ffd100f1ca870 with few changes.

aabawer:
Thanks.
Good catch for the leak.

PR #19, opened 2019-06-04
Parse paths as unicode strings or bytes

aabawer:
We would like to preserve Python 2 handle of paths as either unicode (ascii) strings or bytes.
We apply path converter for accepting PyObject and detecting if its either given as bytes or if its given as unicode. In the latter case we convert the unicode string to its bytes encoding and treat path internally as bytes throughout sanlock API. Tests are added and modified accordingly.

### DIFF
--- a/python/sanlock.c
+++ b/python/sanlock.c
@@ -659,9 +659,9 @@ Sector can be one of (512, 4096).");
 static PyObject *
 py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 {
-    int rv, rs_len, sector = SECTOR_SIZE_512;
+    int rv = -1, rs_len, sector = SECTOR_SIZE_512;
     long align = ALIGNMENT_1M;
-    const char *path;
+    PyObject *path = NULL;
     struct sanlk_resource *rs;
     PyObject *rs_info = NULL;
 
@@ -681,20 +681,20 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
     rs->num_disks = 1;
 
     /* parse python tuple */
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "s|kli", kwlist,
-        &path, &(rs->disks[0].offset), &align, &sector)) {
-        goto exit_fail;
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "O&|kli", kwlist,
+        pypath_converter, &path, &(rs->disks[0].offset), &align, &sector)) {
+        goto finally;
     }
 
     /* prepare the resource disk path */
-    strncpy(rs->disks[0].path, path, SANLK_PATH_LEN - 1);
+    strncpy(rs->disks[0].path, PyBytes_AsString(path), SANLK_PATH_LEN - 1);
 
     /* set alignment/sector flags */
     if (add_align_flag(align, &rs->flags) == -1)
-        goto exit_fail;
+        goto finally;
 
     if (add_sector_flag(sector, &rs->flags) == -1)
-        goto exit_fail;
+        goto finally;
 
     /* read sanlock resource (gil disabled) */
     Py_BEGIN_ALLOW_THREADS
@@ -703,7 +703,7 @@ py_read_resource(PyObject *self __unused, PyObject *args, PyObject *keywds)
 
     if (rv != 0) {
         __set_exception(rv, "Sanlock resource read failure");
-        goto exit_fail;
+        goto finally;
     }
 
     /* prepare the dictionary holding the information */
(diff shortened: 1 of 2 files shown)


python/sanlock.c
```diff
@@ -181,31 +182,33 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
-        if (!PyArg_ParseTuple(disk, "sK", &path, &offset)) {
+        if (!PyArg_ParseTuple(disk, "O&K", pypath_converter, &path, &offset)) {
             /* Override the error since it confusing in this context. */
             set_error(PyExc_ValueError, "Invalid disk %s", disk);
             goto exit_fail;
         }
 
-        strncpy(res->disks[i].path, path, SANLK_PATH_LEN - 1);
+        strncpy(res->disks[i].path, PyBytes_AsString(path), SANLK_PATH_LEN - 1);
+        Py_DECREF(path);
```
nirs: We should use Py_CLEAR(path) here to set the path to NULL immediately after decrefing it. The current code works, but if someone will remove line 188 setting the path to NULL, we will crash on parsing failure.

python/sanlock.c
```diff
@@ -181,31 +182,33 @@ __parse_resource(PyObject *obj, struct sanlk_resource **res_ret)
...
         res->disks[i].offset = offset;
     }
 
     *res_ret = res;
     return 0;
 
 exit_fail:
+    Py_XDECREF(path);
```
nirs: Cleaning up path both globally and both once per loop is complicated and fragile. It is better to extract the code for parsing single (path, offset) item to a helper function that will hide the temporary objects  and have simple cleanup.

So should have something like:
```
parse_disks()
{
    allocate res...

        disk = PyList_GetItem(obj, i);
        if (parse_disks_item(disk, &res->disks[i]))
            goto exit_fail;
    }
    
exit_fail:
    ...
}
```

nirs (changes requested):
Reviewed up to python: Apply pypath_converter to parse_resource util function.

aabawer:
Updated the PR, Thanks.

python/sanlock.c
```diff
@@ -159,6 +159,33 @@ pystring_as_cstring(PyObject *obj)
...
+
+    strncpy(res_disk->path, PyBytes_AsString(path), SANLK_PATH_LEN - 1);
+    res_disk->offset = offset;
+    rv = 1;
+
+exit_fail:
+    Py_XDECREF(path);
+    return rv;
+}
+
```
nirs, line 188: Looks great, so much simpler now.

tests/python_test.py
```diff
@@ -600,8 +617,10 @@ def test_init_lockspace_parse_args(no_sanlock_daemon, name):
 
 
 @pytest.mark.parametrize("name", LOCKSPACE_OR_RESOURCE_NAMES)
-def test_init_resource_parse_args(no_sanlock_daemon, name):
-    disks = [("path", 0)]
+@pytest.mark.parametrize("filename,encoding", FILE_NAMES_NO_XFAILS)
+def test_init_resource_parse_args(no_sanlock_daemon, name, filename, encoding):
+    path = util.generate_path("/tmp/", filename, encoding)
+    disks = [(path, 0)]
     with raises_sanlock_errno(errno.ENOENT):
```
nirs, line 618: The tests changes should be separate patch. You are missing here improvement of the tests and changing the code. This is very likely to fail creating tests that do not work. I will have more confidence when I see the tests fail - before this change.

nirs (changes requested):
Reviewed again up to:
python: Apply pypath_converter to parse_resource util function

aabawer:
this is what we talked about before this PR - to change each test as part
of each relevant patch

python/sanlock.c
aabawer: should be "finally" since we handle success as well

tests/python_test.py
aabawer: this is what we talked about before this PR - to change each test as part
of each relevant patch

aabawer: unless you meant the case where the args parse test is added for the first time.

nirs: We discussed that the patch fixing a failing test should enable the failing test, but here you modified the test and this make the code harder to review. If you think we need to modify the tests by adding parameters for path, why not submit this change as a separate commit before you modify the code?

nirs: The only test change that should be included in this commit is replacing:

    @pytest.mark.parametrize("filename,encoding", FILE_NAMES)

With:

    @pytest.mark.parametrize("filename,encoding", FILE_NAMES_NO_XFAILS)

aabawer:
i think we are talking about the same stuff.

aabawer:
Pushed update to PR for review.

nirs (commented):
Thanks, looks good.

nirs (approved):

PR #21, opened 2020-05-02
Python: Remove init_{lockspace, resource} API

aabawer:
init_{lockspace, resource} API was deprecated by write_{lockspace, resource} API
and only kept for legacy py2 code. This is time to removed it as py2 is now officially retired.


nirs:
Merged in master.

